Issue #31, opened 2014-11-09
Document 'refresh-always' 


nsimmons:
Just noticed that we have no documentation or example on how to use 'refresh-always'.

We need to add something about this.

PR #53, opened 2014-12-19
Add tg-static, fix up refresh-always

qq99:
Fixes https://github.com/Shopify/turbograft/issues/31

With the `tg-static` attribute decorating a node, we can make sure that this node is not replaced during a fullpage refresh.

Though, if you were to refresh the page at a higher level -- e.g., refreshing an ancestor of the tg-static, the static aspect is no longer obeyed and it is replaced.

I also fixed up `refresh-always` key to literally always refresh the node annotated.  Currently, it was refreshing only if you did a `Page.refresh(onlyKeys: [...])` with some keys, which is pretty bad naming for something called `refresh-always` lol

Potential TODO:
- in another PR, migrate `refresh-always` to `tg-always` or something like that?

cc @tmlayton @alexaitken: the feature we need for a static navigation
cc @tylermercier @celsodantas @nsimmons @patrickdonovan @kurtfunai for thoughts and input on adding a new feature to TurboGraft

### DIFF
--- a/test/browser/full_page_refresh_test.rb
+++ b/test/browser/full_page_refresh_test.rb
@@ -56,7 +56,17 @@ class FullPageRefreshTest < ActionDispatch::IntegrationTest
     page.fill_in 'badgeinput', :with => 'tg-static innards'
     click_link "Perform a full page refresh"
     assert_equal "tg-static innards", find_field("badgeinput").value
-    click_link "Perform a partial page refresh and refresh the navigation"
+    click_link "Perform a partial page refresh and refresh the navigation section"
     assert_equal "", find_field("badgeinput").value
+  end
+
+  test "always-refresh will always refresh the annotated nodes, regardless of refresh type" do
+    page.fill_in 'badgeinput2', :with => 'some innards'
+    click_link "Perform a full page refresh"
+    assert_equal "", find_field("badgeinput2").value
+
+    page.fill_in 'badgeinput2', :with => 'some innards'
+    click_link "Perform a partial page refresh and refresh the navigation section"
+    assert_equal "", find_field("badgeinput2").value
   end
 end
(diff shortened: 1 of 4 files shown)


qq99:
:ship: ?

kurtfunai:
:+1:

test/browser/full_page_refresh_test.rb
```diff
@@ -51,4 +51,22 @@ class FullPageRefreshTest < ActionDispatch::IntegrationTest
     page.evaluate_script('window.history.back()')
     assert_equal "Sample Turbograft Application", page.title
   end
+
+  test "tg-static preserves client-side state of innards on partial refresh, and replaces contents on full refresh" do
+    page.fill_in 'badgeinput', :with => 'tg-static innards'
+    click_link "Perform a full page refresh"
+    assert_equal "tg-static innards", find_field("badgeinput").value
+    click_link "Perform a partial page refresh and refresh the navigation section"
+    assert_equal "", find_field("badgeinput").value
```
celsodantas: Is this test correct?
According to line `57` and `58` you do a fullpage refresh and the value is kept. 
On `59` and `60` you do a partial and the value is replaced.

The test title is saying the opposite.

qq99: I think this is correct, just a bit nuanced:

When doing a full page refresh, the stuff inside a `tg-static` (the badgeinput innards) should remain and not be swapped out
But when explicitly replacing the element (`"refresh the navigation section"`) we should see a new DOM element appear (hence the badgeinput now being `''`)

qq99: o I see, @celsodantas clarified offline that the test name is confusing, updated

celsodantas: =D <3

alexaitken:
:clap:

qq99:
https://github.com/Shopify/turbograft/releases/tag/v0.1.5

PR #54, opened 2014-12-19
guard parentNode being null when walking DOM

richardmonette:
While walking the DOM it is possible that parentNode can become null. As a result, the not null check at line 53 may not guard against all null cases. This can be fixed by moving the check into the while loop conditions.

@tylermercier @celsodantas @nsimmons @patrickdonovan @kurtfunai @qq99

### DIFF
--- a/lib/assets/javascripts/turbograft/initializers.coffee
+++ b/lib/assets/javascripts/turbograft/initializers.coffee
@@ -50,9 +50,8 @@ TurboGraft.handlers.remoteFormHandler = (ev) ->
 documentListenerForButtons = (eventType, handler, useCapture = false) ->
   document.addEventListener eventType, (ev) ->
     target = ev.target
-    return if !target
 
-    while target != document && (typeof target != "undefined")
+    while target != document && (typeof target != "undefined") && (target != null)
       if target.nodeName == "A" || target.nodeName == "BUTTON"
         isNodeDisabled = nodeIsDisabled(target)
         ev.preventDefault() if isNodeDisabled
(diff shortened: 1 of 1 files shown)


lib/assets/javascripts/turbograft/initializers.coffee
```diff
@@ -50,9 +50,8 @@ TurboGraft.handlers.remoteFormHandler = (ev) ->
 documentListenerForButtons = (eventType, handler, useCapture = false) ->
   document.addEventListener eventType, (ev) ->
     target = ev.target
-    return if !target
 
-    while target != document && (typeof target != "undefined")
+    while target != document && (typeof target != "undefined") && (target != null)
```
richardmonette: My rationale was that `target != null` will cause an error is target is not defined, whereas (I think anyways) that `typeof target != "undefined"` will return false even if target is not defined. Therefore, if target was undefined, or null, the checked would be false, an not in an error state. That being said, in this context I don't see how target would be undefined, so I agree with your solution. Updated in latest commit, thanks!

richardmonette:
@tylermercier @qq99 I'm going to have to refresh my memory after the holidays, but I'll work on getting a test for this.

lib/assets/javascripts/turbograft/initializers.coffee
pushrax: Just for the future, `typeof undefined == 'undefined'` and `typeof something_never_declared == 'undefined'`

qq99:
Can we ship this as is?

nsimmons:
I would :shipit:

lib/assets/javascripts/turbograft/initializers.coffee
```diff
@@ -50,9 +50,8 @@ TurboGraft.handlers.remoteFormHandler = (ev) ->
 documentListenerForButtons = (eventType, handler, useCapture = false) ->
   document.addEventListener eventType, (ev) ->
     target = ev.target
-    return if !target
 
-    while target != document && (typeof target != "undefined")
+    while target != document && target?
```
qq99: wonder if this can be just `target` and not `target?`

Issue #16, opened 2014-09-26
Investigate why we had to downgrade xhr_headers.rb


qq99:
https://github.com/Shopify/turbograft/pull/15

qq99:
@karlhungus can we consider this investigated?  Could you do a little write-up on this?

karlhungus:
>@karlhungus can we consider this investigated? Could you do a little write-up on this?

I think so

Pasted from @Thibaut's comment [here](https://github.com/Shopify/shopify/issues/35973#issuecomment-69348329):

>So the flow of events is:
>- non-Turbolinks request to Rails redirects to `cdn.shopify.com` (or whatever)
>- because of the missing check, `session[:_turbolinks_redirect_to]` is set to `cdn.shopify.com`
>- next (unrelated) Turbolinks request to Rails pops the session and responds with `X-XHR-Redirected-To: cdn.shopify.com`
>- Turbolinks gets the response back and tries to set `cdn.shopify.com` as the current window location, >which isn't allowed (and even if it were, would be wrong)
>
>Only requests coming from Turbolinks should set `session[:_turbolinks_redirect_to]`.

TLDR: we shouldn't have been setting  `session[:_turbolinks_redirect_to]` for every request.

Feel free to reopen if this isn't sufficient :)

qq99:
Thanks :)

PR #56, opened 2015-01-09
don't set session if redirected not set

karlhungus:
@Thibaut @qq99 

##### Problem 

I found an issue in our usage of this where redirecting on location.replace would set `session[:_turbolinks_redirect_to]` without headers. This would cause subsequent requests to get redirected when they shouldn't

https://github.com/Shopify/shopify/issues/35973#issuecomment-69338390

##### Solution

Update to latest version of `xhr_headers`: https://github.com/rails/turbolinks/blob/b7553314d82f9907ea6da57361bb4fe9e956eaf6/lib/turbolinks/xhr_headers.rb#L28. I followed the history of xhr_headers, and all changes seem reasonable. 

I have yet to double check this fixes my actual problem :).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,13 @@
 language: ruby
 rvm:
-  - "2.0.0"
+  - '2.0.0'
+  - '2.1.2'
+  - '2.2.0'
 before_script:
   - sh -e /etc/init.d/xvfb start
   - bundle install
 env:
-  - DISPLAY=":99.0"
+  - DISPLAY=':99.0' RAILS_VERSION='3.2.0'
+  - DISPLAY=':99.0'  RAILS_VERSION='4.1.0'
+  - DISPLAY=':99.0'  RAILS_VERSION='4.2.0'
 script: bundle exec rake ci
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -20,6 +20,7 @@ class TurboGraft.Remote
     xhr.setRequestHeader('X-Requested-With', 'XMLHttpRequest')
     xhr.setRequestHeader('Accept', 'text/html, application/xhtml+xml, application/xml')
     xhr.setRequestHeader("Content-Type", @contentType) if @contentType
+    xhr.setRequestHeader 'X-XHR-Referer', document.location.href
 
     csrfToken = CSRFToken.get().token
     xhr.setRequestHeader('X-CSRF-Token', csrfToken) if csrfToken
--- a/lib/turbograft/xhr_headers.rb
+++ b/lib/turbograft/xhr_headers.rb
@@ -8,39 +8,40 @@ module TurboGraft
   # request, where it will be used to set the X-XHR-Redirected-To response header.  The
   # Turbolinks script will detect the header and use replaceState to reflect the redirected
   # url.
+
   module XHRHeaders
     extend ActiveSupport::Concern
 
-    included do
-      alias_method_chain :_compute_redirect_to_location, :xhr_referer
+    def _compute_redirect_to_location(*args)
+      options, request = _normalize_redirect_params(args)
+
+      store_for_turbolinks begin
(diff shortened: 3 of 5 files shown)


Thibaut:
cc @arthurnn 
will look at this around noon

Thibaut:
This needs a good :tophat: but otherwise :+1:.

qq99:
want to make a PR ref'ing this SHA ?

karlhungus:
> want to make a PR ref'ing this SHA ?

Will do, 

@qq99 any notion why `PartialPageRefreshTest#test_when_I_use_an_XHR_and_POST_to_an_endpoint_that_returns_me_a_302,_I_should_see_the_URL_reflecting_that_redirect_too` wouldn't be setting `X-XHR-Referer` -- I put breakpoints in ['tubolinks.coffee'](https://github.com/Shopify/turbograft/blob/098ead7d84ff22b9f4350346e292f99e3ab55f33/lib/assets/javascripts/turbograft/turbolinks.coffee#L81) but they weren't triggered

qq99:
Is `fetchReplacement` being called?

kurtfunai:
Just going to mention that this seems related to issue #16 / PR #15

karlhungus:
@qq99 noted:
https://github.com/Shopify/turbograft/blob/098ead7d84ff22b9f4350346e292f99e3ab55f33/lib/assets/javascripts/turbograft/remote.coffee#L22
we need to set header in here too I think and then `tg-remote` should be OK

karlhungus:
Fixes: https://github.com/Shopify/shopify/issues/35973

arthurnn:
We should test this repo against multiple rails versions! we patch rails it self, and try to be compatible with multiple versions, so we need to tests those!

karlhungus:
>We should test this repo against multiple rails versions!

I'll add that too

karlhungus:
re-ping @arthurnn @Thibaut @qq99 /cc @Shopify/admin 

This adds the multiple rails versions (kind of guessed here), fixes the test by adding the header forceably. Updates tg-remote to make use of the referer (not sure if that was ok).

If this is all-right, i'll merge https://github.com/Shopify/shopify/pull/36047 if things go awry then i'll revert.

qq99:
looks legit :+1:

arthurnn:
:+1:

.travis.yml
```diff
@@ -1,9 +1,12 @@
 language: ruby
 rvm:
-  - "2.0.0"
+  - '2.0.0'
+  - '2.1.2'
```
Thibaut: `2.2.0` ?

.travis.yml
```diff
@@ -1,9 +1,12 @@
...
 before_script:
   - sh -e /etc/init.d/xvfb start
   - bundle install
 env:
-  - DISPLAY=":99.0"
+  - DISPLAY=':99.0' RAILS_VERSION='3.1.0'
+  - DISPLAY=':99.0' RAILS_VERSION='3.2.0'
+  - DISPLAY=':99.0'  RAILS_VERSION='4.1.0'
```
Thibaut: `4.2.0`?

karlhungus:
Any rails version/ruby versions i should remove (with all of these builds take a long time and are a bit more flakey)?

kurtfunai:
@karlhungus I might be wrong, but I think that rails 3.1 is no longer supported?

Found an old article, http://weblog.rubyonrails.org/2013/2/24/maintenance-policy-for-ruby-on-rails/
Security issues:
>Severe security issues:
>For severe security issues we will provide new versions as above, and also the last major release series will receive patches and new versions. The classification of the security issue is judged by the core team.
>Currently included series: 3.2.z, 3.1.z, 2.3.z
>After the Rails 4 release: 4.0.z, 3.2.z

karlhungus:
Thanks @kurtfunai , 3.1 removed :). Cool to merge this -- then i'll create 0.1.6?

qq99:
merge :+1:

PR #57, opened 2015-01-13
Fixed a bug with `page:load` event not having proper event.data

qq99:
cc @karlhungus 

Bug:

Listening to the TurboGraft emitted event `page:load`, as of `0.1.5`, does not have a list of new nodes in `event.data` as it should have

Explanation:

- `refreshNodes: ->` returns a list of nodes that were refreshed
- When I refactored `refreshNodesWithKeys: ->` into 2 separate fns, `refreshRefreshAlwaysNodes: ->` and `refreshNodesWithKeys: ->` (it previously did both of these in 1 function), I forgot to return `refreshedNodes` properly
- `changePage` returns a list of refreshed nodes or void, and this event is fired:  `triggerEvent 'page:load', nodes`
- ^ previously, `nodes` in the event triggered above was always `null`, now it is something!

Effect:

If you had a listener for `page:load` in your app and were expecting to see a list of nodes you would never see any nodes.  Say, you were using `Twine.refresh(newNodes)`, you would essentially never refresh the new nodes!

### DIFF
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -119,9 +119,11 @@ class window.Turbolinks
   changePage = (title, body, csrfToken, runScripts, partialReplace, onlyKeys = [], exceptKeys = []) ->
     document.title = title if title
 
-    refreshRefreshAlwaysNodes(body)
+    refreshedNodes = []
+    refreshedNodes = refreshRefreshAlwaysNodes(body)
     if onlyKeys.length
-      return refreshNodesWithKeys(onlyKeys, body)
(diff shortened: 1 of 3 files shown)


karlhungus:
Verified that admin/settings/payments still works and the test that caught it.

karlhungus:
:fries:

qq99:
Added a test as well, we had 0 around `page:load` event firing and its parameters
cc @tylermercier @kurtfunai @patrickdonovan

patrickdonovan:
Makes sense :shipit:

karlhungus:
@qq99, I've just been manually rebuilding :( ...

qq99:
I increased timeout to 7s and it passed

PR #58, opened 2015-01-14
Prevent refresh-always from destroying nodes

qq99:
When replacing nodes, it seems like `existingNode.parentNode.replaceChild(newNode, existingNode);` will move `newNode` out of any current tree its in, and into the tree where `existingNode` used to sit.

`0.1.5` made `refresh-always` call way more often, not just if there were `onlyKeys` to refresh.  Thus, this line of code was being run & and any `existingNode` that had `refresh-always` on it was being moved from the server-fresh `body` and being placed into `document.body`.  But!-- since it's a full page refresh, the next step is replacing `document.body` with server-fresh `body`, which now lacks this node, leading to the appearance of the `refresh-always` nodes disappearing.

The solution I suggest is to make a deep clone of the node when replacing.

cc @karlhungus @patrickdonovan @tylermercier @kurtfunai 

Crazy enough, we didn't catch this in our example app because the only `refresh-always` node was inside a `tg-static` node, which masked its disappearance on a full refresh :(  I've added an example to the app now and a test against it

This will become `v0.1.8`

Related: https://github.com/Shopify/shopify/pull/36344

### DIFF
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -163,6 +163,7 @@ class window.Turbolinks
         throw new Error "Turbolinks refresh: Refresh key elements must have an id."
 
       if newNode = body.querySelector("##{ nodeId }")
+        newNode = newNode.cloneNode(true)
         existingNode.parentNode.replaceChild(newNode, existingNode)
 
         if newNode.nodeName == 'SCRIPT' && newNode.getAttribute("data-turbolinks-eval") != "false"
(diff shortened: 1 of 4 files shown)


karlhungus:
>  leading to the appearance of the refresh-always nodes disappearing.

Heh :)

Any concerns about deep cloning that node being expensive?

Other than that :fries:

celsodantas:
Same concern. Looks good.
Failing test, not sure if related.

qq99:
Legit test failure, reformulated code slightly to make it pass (it was trying to execute script of old node, not new node)

qq99:
I think `cloneNode` will not present any speed penalties

qq99:
Tried profiling things, seems negligible

kurtfunai:
I think this looks good - Nice fix! :smile:

celsodantas:
:+1:

qq99:
gonna ship!

PR #59, opened 2015-01-14
Don't automatically add _method to FormData

DrewMartin:
This will let us properly https://github.com/Shopify/shopify/issues/36350

We can't set `tg-remote` on the products form because the form's `method="POST"` attribute that rails is setting overwrites the `method="PATCH"` hidden field within the form. We should just not set the submission type based on form attributes, and instead use a `_method` hidden field. Ideally we'd look at what's already in `formData` but we can't because FormData is stored in native code and we don't have access.

@qq99 @celsodantas

### DIFF
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -58,9 +58,7 @@ class TurboGraft.Remote
     else
       formData = ''
 
-    if formData instanceof FormData
-      formData.append("_method", @opts.httpRequestType) if @opts.httpRequestType
-    else
+    if formData not instanceof FormData
       @contentType = "application/x-www-form-urlencoded; charset=UTF-8"
       formData = @formAppend(formData, "_method", @opts.httpRequestType) if formData.indexOf("_method") == -1 && @opts.httpRequestType && @actualRequestType != 'GET'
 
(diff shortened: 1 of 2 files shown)


celsodantas:
no test to update? Can we add one to avoid someone putting this back?

I'm totally ok with this change.
cc @karlhungus @kurtfunai for more eyes

qq99:
I don't think there was a way to test FormData since it's so black boxy lol

celsodantas:
haha TRUE! haha
nvm then.

qq99:
Maybe can do a controller test?

DrewMartin:
It would have to be a browser test, maybe that looks at the post data? Not sure how to even approach that.

celsodantas:
what if we check it on the server? Not a js test.

DrewMartin:
We'd still need to run JS code

qq99:
If you make a method that only responds to `PATCH` and you `refresh-on-success` some part of `page.html.erb`, that might be one way to go about it?  Perhaps in the section with the `tg-remote forms`

qq99:
Here's one that you can perhaps re-purpose:
```
<form tg-remote action="/pages/put_foo" method="put" refresh-on-success="foo-results-put" refresh-on-error="foo-results-put">
  <input id="fooput" name="fooput" type="text">
  <button type="submit">Submit tg-remote PUT</button>
</form>
```
by changing the type to `file`?

DrewMartin:
Added a test

qq99:
:+1:

celsodantas:
good to me. Not sure how we mock stuff in Turbograft. If we don't, the test looks good to me.

qq99:
I think that's solid for mocking window objects

celsodantas:
:sheep: :dash:

DrewMartin:
I keep getting test failures in something completely unrelated :/

qq99:
It's pretty flaky with the 5 builds going on

DrewMartin:
Yay green!

PR #60, opened 2015-01-19
fixing autofocus for fullpage refresh

celsodantas:
## Problem

When we refresh the page using turbograft, `autofocus` does not work, only works if you refresh the page.

This was fixed in Turbolinks a while ago, but the fix happened after we branched from it.

## Solution

I'm just copying the same solution from Turbolinks: https://github.com/rails/turbolinks/commit/3aab7249bb0b695b49f2a31493653f0dce87729c
The comment on Turbolinks that the problem is only happening on Firefox is wrong. It also happens on other browsers as well.

I'm focusing only on `fullpage` refresh. The reason is because I don't see any use case to also do that on partial page refresh.

reviews @kurtfunai @qq99 @karlhungus  @patrickdonovan

### DIFF
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -104,7 +104,6 @@ class window.Turbolinks
   @loadPage: (url, xhr, partialReplace = false, onLoadFunction = (->), replaceContents = [], replaceAllExcept = []) ->
     triggerEvent 'page:receive'
 
-
     if doc = processResponse(xhr, partialReplace)
       reflectNewUrl url
       nodes = changePage(extractTitleAndBody(doc)..., partialReplace, replaceContents, replaceAllExcept)
@@ -133,6 +132,7 @@ class window.Turbolinks
       triggerEvent 'page:before-replace'
       document.documentElement.replaceChild body, document.body
       CSRFToken.update csrfToken if csrfToken?
+      setAutofocusElement()
       executeScriptTags() if runScripts
       currentState = window.history.state
       triggerEvent 'page:change'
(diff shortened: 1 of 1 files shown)


karlhungus:
Should we add a test? Other than that :fries:

kurtfunai:
:+1: for this fix - I was running into this problem as well.

patrickdonovan:
:+1:

tylermercier:
+1 for adding a test

qq99:
Sounds pretty solid, if it turns out we need it on partial-page refresh, we can do so

I'm not entirely sure how you'd test this ;p

karlhungus:
Wouldn't you only have to refresh and get a page that had an autofocus element, then verify that was autofocused? Based on the comment in the turbolinks commit , it looks like it's only firefox...

qq99:
Yeah I just wonder how good Capybara is at detecting focus reliably

kurtfunai:
@qq99 I was talking to @celsodantas about this, but I am presently using turbograft to render forms using partial-page replacements in my project. 
Might be nice if it worked for partial page replacements as well! :smile_cat:

celsodantas:
can't make the tests pass. It works locally on Chrome. Guess I can't test this in Phantom. =/

tylermercier:
>wow phantomjs is the new IE

karlhungus:
> Guess I can't test this in Phantom

I've had similar experiences, i'm fine with no test for this then.

Was i wrong about it being a firefox only problem?

celsodantas:
Yeah @karlhungus , the problem happens in other browsers as well. The comment on the Rails code is wrong. Here is more people having the same issue with Chrome: https://github.com/rails/turbolinks/issues/365 (including myself)

celsodantas:
Removed the tests and kept the partial page refresh autofocus fix.
can I get another :eyes: ?

kurtfunai:
Looks good to me!

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -155,6 +157,18 @@ class window.Turbolinks
 
     return matchingNodes
 
+  anyAutofocusElement = (nodes) ->
+    for node in nodes
+      if node.querySelectorAll('input[autofocus], textarea[autofocus]').length > 0
+        return true
```
qq99: so I forget my coffeescript, does this return from function or is it returning `true` to the comprehension of `for node in nodes`?

tylermercier: The former, it will break out of the `anyAutofocusElement` function early.

qq99:
LGTM

PR #61, opened 2015-01-19
Avoid returning stale nodes by doing replace in one shot, as it was performed previously

qq99:
cc @patrickdonovan @DrewMartin @tylermercier @celsodantas @kurtfunai 

```coffee
document.addEventListener 'page:load', (event) ->
  reset(event.data)
```

This should fix the bug we saw earlier where the returned `nodes` (`event.data` in above listener) during partial-refresh had a stale node inside it.  Ie, nodes that are not even attached to the `window.body`.

An example problematic set of returned nodes from a partial-replace in `0.1.9` is:

- `foo` (a `refresh-always` node)
- `bar` (a `refresh="something"` node that is ancestor of `foo`)

`foo` is first element in array b/c of `refresh-always` processing.  Then `bar` is next because of `onlyKeys` processing.

However, since `bar` encapsulates `foo`, `foo` is stale and no longer in the DOM, but it is returned to any listeners of the event.  

`0.1.5` created this problem because it split the refreshOnSuccess strategy into 1 phase for `refreshAlways` nodes and a second phase for `refreshKeys` nodes.  The solution in this PR is to do a single refresh, but with a larger set of nodes (more akin to `<=0.1.4`).  In the case where we are not doing `onlyKeys`, we still refresh the `refresh-always` so we end up with `refresh-always` always firing (which was the intended result of `0.1.6`).

The `parentIsRefreshing` function guards against this staleness, but in order for it to work correctly, the call to `refreshNodes` must have the entire known set of nodes that are being refreshed and cannot be split into 2 calls of 2 different sets

### DIFF
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -119,12 +119,11 @@ class window.Turbolinks
   changePage = (title, body, csrfToken, runScripts, partialReplace, onlyKeys = [], exceptKeys = []) ->
     document.title = title if title
 
-    refreshedNodes = []
-    refreshedNodes = refreshRefreshAlwaysNodes(body)
     if onlyKeys.length
-      refreshedNodes = refreshedNodes.concat refreshNodesWithKeys(onlyKeys, body)
-      return refreshedNodes
+      nodesToRefresh = [].concat(getNodesWithRefreshAlways(), getNodesMatchingRefreshKeys(onlyKeys))
+      return refreshNodes(nodesToRefresh, body)
     else
+      refreshNodes(getNodesWithRefreshAlways(), body)
       persistStaticElements(body)
       if exceptKeys.length
         refreshAllExceptWithKeys(exceptKeys, body)
@@ -140,6 +139,21 @@ class window.Turbolinks
       triggerEvent 'page:update'
 
     return
+
+  getNodesMatchingRefreshKeys = (keys) ->
+    matchingNodes = []
+    for key in keys
+      for node in document.querySelectorAll("[refresh=#{key}]")
+        matchingNodes.push(node)
+
+    return matchingNodes
+
+  getNodesWithRefreshAlways = ->
+    matchingNodes = []
+    for node in document.querySelectorAll("[refresh-always]")
+      matchingNodes.push(node)
+
+    return matchingNodes
 
   deleteRefreshNeverNodes = (body) ->
     for node in body.querySelectorAll('[refresh-never]')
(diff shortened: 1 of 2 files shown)


tylermercier:
Tests :trollface:

qq99:
Will add some tests in the morning
@DrewMartin this patch will make that event you were using only emit once (previously you mentioned it was emitting twice, but only once with the node you wanted)

qq99:
Added 2 tests, last one demonstrates that this bug has been fixed, first one is more a sanity check that `page:before-partial-replace` is only called once

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -141,6 +140,21 @@ class window.Turbolinks
 
     return
 
+  getNodesMatchingRefreshKeys = (keys) ->
+    matchingNodes = []
+    for key in keys
+      for node in document.querySelectorAll("[refresh=#{key}]")
+        matchingNodes.push(node)
```
tylermercier: How slow is this?

qq99: yeah, previously was finding them all & replacing, now just finding them and returning them (smaller methods ftw)

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -141,6 +140,21 @@ class window.Turbolinks
...
+
+    return matchingNodes
+
+  getNodesWithRefreshAlways = ->
+    matchingNodes = []
+    for node in document.querySelectorAll("[refresh-always]")
+      matchingNodes.push(node)
+
+    return matchingNodes
```
tylermercier: We could use `map` here instead.

qq99: Ideally this `NodeList` would have a proper `toArray()` ><  maybe I can find one

qq99: https://developer.mozilla.org/en/docs/Web/API/NodeList I don't think we can use map lol

qq99: How do you feel about:

```
var div_list = document.querySelectorAll('divs'); // returns NodeList
var div_array = Array.prototype.slice.call(div_list); // converts NodeList to Array
```

tylermercier: That is :hankey:

tylermercier:
```
nodes = document.querySelectorAll("[refresh-always]")
Array.prototype.slice.call(nodes).map ...
```

qq99:
How do you feel about :ship: on :green_apple:  for `0.2.0`  Or are we naming it `0.1.10`?

patrickdonovan:
:tophat:ed in shopify. I vote for `0.1.10` but don't feel strongly. :boat:

karlhungus:
fwiw we usually use this: http://semver.org/

```
Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible manner, and
PATCH version when you make backwards-compatible bug fixes.
Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
```

PR #64, opened 2015-01-22
Enable turbograft on ios chrome

### LINKED ISSUE #62: Enable on ios chrome
We explicitly disable turbograft on chrome for ios ([here](https://github.com/Shopify/turbograft/blob/master/lib/assets/javascripts/turbograft/turbolinks.coffee#L22-L23)) because [that's what turbolinks does](https://github.com/rails/turbolinks/blob/e5670fe2caf333a991a4b0949f762ad3e7aa8611/lib/assets/javascripts/turbolinks.js.coffee#L526-L527). The issue where this was brought up was https://github.com/rails/turbolinks/issues/82.

I don't think this is the case any more. At least, experimenting with it on my iphone and looking at the values in chrome for ios, I'm not seeing the order issue that initially caused this. I think we might be able to remove the check.

Thoughts? @kristianpd @qq99 @patrickdonovan @nsimmons @pushrax @angelini 

This would fix https://github.com/Shopify/shopify/issues/29477

DrewMartin:
Closes #62

Originally done because that's what turbolinks does (rails/turbolinks#82) but it looks like either our version of turbolinks or ios chrome doesn't suffer from this bug any more. This will fix https://github.com/Shopify/shopify/issues/29477

@qq99 @patrickdonovan @celsodantas @maartenvg

### DIFF
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -19,9 +19,6 @@ historyStateIsDefined =
 browserSupportsPushState =
   window.history and window.history.pushState and window.history.replaceState and historyStateIsDefined
 
-browserIsntBuggy =
-  !navigator.userAgent.match /CriOS\//
-
 window.triggerEvent = (name, data) ->
   event = document.createEvent 'Events'
   event.data = data if data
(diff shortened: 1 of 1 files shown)


qq99:
:+1:

DrewMartin:
Is there anyone else I should ping? @qq99 who's maintaining this? Just you?

karlhungus:
I don't know how to test this, but otherwise the change looks fine.

:fries:

qq99:
@tylermercier @kurtfunai

DrewMartin:
@karlhungus you'd need an iphone (or ipad I guess) and follow the steps in https://vault.shopify.com/Local-Mobile-Device-Testing

celsodantas:
:+1: to me

PR #65, opened 2015-02-02
add refresh on success except logic

richardmonette:
Adding `full-refresh-on-success-except` which will allow for selectively omitting parts of the page when refreshing on success.

review @qq99 @karlhungus

### DIFF
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -8,6 +8,7 @@ class TurboGraft.Remote
     @formData = @createPayload(form)
 
     @refreshOnSuccess       = @opts.refreshOnSuccess.split(" ")       if @opts.refreshOnSuccess
+    @refreshOnSuccessExcept = @opts.refreshOnSuccessExcept.split(" ") if @opts.refreshOnSuccessExcept
     @refreshOnError         = @opts.refreshOnError.split(" ")         if @opts.refreshOnError
     @refreshOnErrorExcept   = @opts.refreshOnErrorExcept.split(" ")   if @opts.refreshOnErrorExcept
 
@@ -102,6 +103,10 @@ class TurboGraft.Remote
         Page.refresh
           response: xhr
           onlyKeys: @refreshOnSuccess
+      else if @refreshOnSuccessExcept
+        Page.refresh
+          response: xhr
+          exceptKeys: @refreshOnSuccessExcept
       else
         Page.refresh
           response: xhr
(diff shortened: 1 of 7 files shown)


lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -102,9 +103,13 @@ class TurboGraft.Remote
         Page.refresh
           response: xhr
           onlyKeys: @refreshOnSuccess
-      else
+      else if @refreshOnSuccessExcept
         Page.refresh
           response: xhr
+          exceptKeys: @refreshOnSuccessExcept
```
qq99: there's no `onlyKeys` for this usage pattern?

qq99: also, where's the `refreshOnErrorExcept` branch?

qq99: nvm, I see it's called `full-refresh-on-success-except` :+1:

test/javascripts/initializers_test.coffee
```diff
@@ -106,10 +112,11 @@ describe 'Initializers', ->
         httpUrl: "somewhere"
         fullRefresh: false
         refreshOnSuccess: "foo"
+        refreshOnSuccessExcept: null
         refreshOnError: null
         refreshOnErrorExcept: null
 
-    it 'passes through null for missing full-refresh-on-error-except', ->
+    it 'respects full-refresh-on-error-except', ->
```
qq99: `full-refresh-on-error-except` is a bit of a misnomer as it can also be a partial refresh on error except:
```
    if @refreshOnError || @refreshOnErrorExcept
      Page.refresh
        response: xhr
        onlyKeys: @refreshOnError
        exceptKeys: @refreshOnErrorExcept
```

richardmonette: @qq99 I think it makes sense to stay consistent with the error case logic. I've updated the code and adjusted the tests slightly.

qq99:
Usage pattern looks solid, can you make sure README is up to date with this info as well?  As well as some of the stipulations we talked about in comments

richardmonette:
Updated the README to reflect the changes

celsodantas:
Nothing stands out. :+1:

richardmonette:
@DrewMartin can I get your thoughts on this too?

lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -114,10 +119,13 @@ class TurboGraft.Remote
       initiator: @initiator
       xhr: xhr
 
-    if @refreshOnError || @refreshOnErrorExcept
```
DrewMartin: Was this a bug?

DrewMartin:
Looks good

lib/assets/javascripts/turbograft/remote.coffee
richardmonette: Yes, although in practice it's unlikely anyone would have noticed. In turbolinks.coffee, in changePage, if there are only keys, then we don't use the expect keys. What that meant, is that if you had both, we never used the expect keys. In most cases, the section which would be refreshed on error would not overlap the section to not refresh on error, so it wouldn't really matter, i.e. by only refreshing the on refresh keys, you would as a side effect not refresh the except keys. My opinion is that by changing this around, into separate if statements, we make the behaviour clearer and more explicit.

PR #66, opened 2015-02-04
filter out inputs which do not have name attribute

richardmonette:
fixes https://github.com/Shopify/shopify/issues/37614

In IE11 it appears that serializing file inputs without data causes issues. This PR adds tg-remote-noserialize which can be used to omit such inputs.

review @qq99 @DrewMartin @maartenvg

### DIFF
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -51,10 +51,13 @@ class TurboGraft.Remote
 
   createPayload: (form) ->
     if form
-      if form.querySelectorAll("[type='file']").length > 0
-        formData = new FormData(form)
+      $cleanForm = $(form).clone()
+      $cleanForm.find('input[tg-remote-noserialize]').remove()
+      cleanForm = $cleanForm[0]
+      if cleanForm.querySelectorAll("[type='file']").length > 0
+        formData = new FormData(cleanForm)
       else # for much smaller payloads
-        formData = @uriEncodeForm(form)
+        formData = @uriEncodeForm(cleanForm)
     else
       formData = ''
 
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -52,7 +52,7 @@ class TurboGraft.Remote
   createPayload: (form) ->
     if form
       $cleanForm = $(form).clone()
-      $cleanForm.find('input[tg-remote-noserialize]').remove()
+      $cleanForm.querySelectorAll('input[tg-remote-noserialize]').remove()
       cleanForm = $cleanForm[0]
       if cleanForm.querySelectorAll("[type='file']").length > 0
         formData = new FormData(cleanForm)
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -51,9 +51,9 @@ class TurboGraft.Remote
 
   createPayload: (form) ->
     if form
-      $cleanForm = $(form).clone()
-      $cleanForm.querySelectorAll('input[tg-remote-noserialize]').remove()
-      cleanForm = $cleanForm[0]
+      cleanForm = form.cloneNode(true)
+      for node in cleanForm.querySelectorAll('input:not([name])')
+        node.parentNode.removeChild(node)
       if cleanForm.querySelectorAll("[type='file']").length > 0
         formData = new FormData(cleanForm)
       else # for much smaller payloads
--- a/test/javascripts/remote_test.coffee
+++ b/test/javascripts/remote_test.coffee
@@ -358,6 +358,18 @@ describe 'Remote', ->
       remote = new TurboGraft.Remote({}, form)
       assert (remote.formData instanceof FormData)
 
+    it 'will not create FormData object if the only input has tg-remote-noserialize', ->
+      form = $("<form><input tg-remote-noserialize type='file' name='foo'></form>")[0]
+
+      remote = new TurboGraft.Remote({}, form)
+      assert.isFalse (remote.formData instanceof FormData)
+
+    it 'will create FormData object but skip any input with tg-remote-noserialize', ->
+      form = $("<form><input type='file' name='foo'><input tg-remote-noserialize type='file' name='bar'></form>")[0]
+
+      remote = new TurboGraft.Remote({}, form)
+      assert (remote.formData instanceof FormData)
+
     it 'will add the _method to the form if supplied in the constructor', ->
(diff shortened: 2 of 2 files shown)


richardmonette:
updated to use querySelectorAll over jquery find and adding tests

lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -51,10 +51,13 @@ class TurboGraft.Remote
 
   createPayload: (form) ->
     if form
-      if form.querySelectorAll("[type='file']").length > 0
-        formData = new FormData(form)
+      $cleanForm = $(form).clone()
```
qq99: this still uses jQuery

qq99: `cleanForm = form.cloneNode(true)` for pure JS API deep clone

lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -51,10 +51,13 @@ class TurboGraft.Remote
...
+      $cleanForm.querySelectorAll('input[tg-remote-noserialize]').remove()
```
qq99:
```coffee
for node in cleanForm.querySelectorAll('input[tg-remote-noserialize]')
  node.parentNode.removeChild(node)
```
to remove a node from its own parent

DrewMartin:
I don't like `tg-remote-noserialize`. There's already two ways to tell that an input shouldn't be serialized: put `disabled` on it, or don't put a name attribute. Can we use one of these to determine that the field shouldn't be serialized.

lib/assets/javascripts/turbograft/remote.coffee
richardmonette: updated

lib/assets/javascripts/turbograft/remote.coffee
richardmonette: Used this suggestion, but combined with @DrewMartin 's suggestion of using the absence of name

nsimmons:
I agree with Drew. There are already ways of preventing a form input form being serialized. There is no need to add another way.

richardmonette:
Updated to use the absence of name as the filter, per Drew's suggestion.

DrewMartin:
:cool:

qq99:
Hmm, but if you `disabled` the file input, it can no longer be used as a method of selecting files, right?

richardmonette:
Not sure about `disabled`, I think the idea is just to use the absence of the `name` attribute.

qq99:
Oh cool, didn't see that comment :+1:

qq99:
:ship: on :green_apple:

maartenvg:
:+1: to the new way.

lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -51,10 +51,13 @@ class TurboGraft.Remote
 
   createPayload: (form) ->
     if form
-      if form.querySelectorAll("[type='file']").length > 0
-        formData = new FormData(form)
+      cleanForm = form.cloneNode(true)
+      for node in cleanForm.querySelectorAll('input:not([name])')
```
qq99: this should be `input:not([name]), textarea:not([name])`

Or alternatively `tg-remote-noserialize` lol

qq99: screen shot 2015-02-04 at 5 43 50 pm

qq99: ^ Red herring, but still a change that should probably happen

> cleanForm.querySelector("#order-note").value
> => "test"

That's the original value from page init.  Perhaps something is just messed up with cloneNode and textareas?

qq99: screen shot 2015-02-04 at 5 49 26 pm

wtf

qq99:
```coffee
node = form.querySelector("#sometextarea")
node.value
# => "foo"
dup = node.cloneNode(true)
dup.value
# => "bar"
```

This is the problem we were seeing :(

DrewMartin:
Yeah I just got to the same point. cloneNode (and $.clone) seem to clone textareas wrong. It looks like the html of a textarea is whatever it was on first page load, but when you clone one the html stays the same and the value gets reverted to whatever the html is

DrewMartin:
Looking at the change set again it should probably be `':input:not([name])'` instead of `'input:not([name])'` but then again this is only for file inputs so either way

qq99:
`:input` matches `textarea` too?

DrewMartin:
Interesting, so selects would probably have the same problem

DrewMartin:
:input matches textarea, select and input

DrewMartin:
and button: http://api.jquery.com/input-selector/

qq99:
Some sources recommend changing node innards (http://stackoverflow.com/questions/21059496/why-dont-select-and-textarea-value-changes-get-updated-upon-cloning-as-with-inp)

Not sure we can faithfully clone things for form submission without doing something like suggested in this link

DrewMartin:
Yeah we'd probably have to modify the form's textareas and selects to have html match value for textarea, and selected attribute be set on the appropriate option on selects.

qq99:
I suppose we can work around needing to remove the elements from a cloned DOM by looping over the ones we do want to serialize (we already do this for uriEncoding)

For FormData, we can loop and I believe there is a `FormData.prototype.append`

DrewMartin:
True, yeah we could just try appending all the nodes ourselves because IE sucks. That would probably be cleaner

DrewMartin:
Does formData need a form argument? I'm not sure if it grabs things like url and method there, on top of inputs. We could probably clone just the form itself and pass it in if that's the case, then iterate over the inputs.

qq99:
so I think https://github.com/Shopify/turbograft/blob/master/lib/assets/javascripts/turbograft/remote.coffee#L58 should be nixed in favour of looping and appending to a `new FormData()` for the most robust fix, the result would look a lot like https://github.com/Shopify/turbograft/blob/master/lib/assets/javascripts/turbograft/remote.coffee#L74-L83

It might be best to rename `uriEncodeForm` to something else.  Then, in there, do the `FormData.append` or `String.append` depending on the presence of `input[type='file']`

qq99:
We should probably yank `0.1.13` at some time

qq99:
Good idea about cloning form :+1: clone it and kill all children for FormData constructor sounds good to me

DrewMartin:
Looks like there's no need to clone the form:
formdata - web api interfaces mdn 2015-02-04 18-05-05

qq99:
Sorry @richardmonette it turns out that my suggestion to cloneNode here was a terrible idea!

PR #67, opened 2015-02-04
create FormData iteratively

DrewMartin:
@qq99 @richardmonette

PR #68, opened 2015-02-05
remove FormData object code path

richardmonette:
There appears to be a bug in IE11 where submitting FormData with empty file inputs causes the request to result in 400 Bad request when it otherwise would have been expected to succeed. Given that this functionality seems to be buggy and that we, Shopify, aren't using this in production anyways, this PR proposes the removal of the FormData object code.

@DrewMartin has also been exploring an option https://github.com/Shopify/turbograft/pull/67 where we continue to support FormData, but given that we aren't using the feature and it's problematic to test this might be a better direction.

@qq99 @DrewMartin @Shopify/admin

### DIFF
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -50,21 +50,9 @@ class TurboGraft.Remote
     @xhr.send(@formData)
 
   createPayload: (form) ->
-    if form
-      cleanForm = form.cloneNode(true)
-      for node in cleanForm.querySelectorAll('input:not([name])')
-        node.parentNode.removeChild(node)
-      if cleanForm.querySelectorAll("[type='file']").length > 0
-        formData = new FormData(cleanForm)
-      else # for much smaller payloads
-        formData = @uriEncodeForm(cleanForm)
-    else
-      formData = ''
-
-    if formData not instanceof FormData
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -50,21 +50,9 @@ class TurboGraft.Remote
...
-      else # for much smaller payloads
-        formData = @uriEncodeForm(cleanForm)
-    else
-      formData = ''
-
-    if formData not instanceof FormData
-      @contentType = "application/x-www-form-urlencoded; charset=UTF-8"
-      formData = @formAppend(formData, "_method", @opts.httpRequestType) if formData.indexOf("_method") == -1 && @opts.httpRequestType && @actualRequestType != 'GET'
-
+    formData = @uriEncodeForm(form)
```
DrewMartin: What happens to file inputs when you encode them this way?

qq99:
I'm still hesitant to remove it since someone may be relying on it outside of Shopify

lib/assets/javascripts/turbograft/remote.coffee
qq99: I feel like it will explode somehow?

richardmonette:
OK, lets find another way.

Issue #73, opened 2015-02-12
Refactor Page.refresh


qq99:
```coffee
Page.refresh
  response: xhr
  onlyKeys: [...]
```
vs
```coffee
Page.refresh
  onlyKeys: [...]
```

Latter does an extra GET, the former uses the response of an XHR, but they're both named `refresh`.  The former should be named `replace` or something

qq99:
@maartenvg I think your PR addressed this issue?

maartenvg:
Not really? It is still called refresh, despite taking different inputs.

qq99:
Gotcha :+1: thanks

TheMallen:
This makes a lot of sense, I'm a little worried that our existing uses in the admin might call Page.refresh with an xhr directly though. I'll take a look. I'd be worried about breaking the api though (especially as long as we have the weird doppelganger version of `Remote` in our core codebase).

PR #74, opened 2015-02-19
Optionally prevent PushState on Page.refresh without XHR

maartenvg:
### Introduction

Currently we have 2 ways of doing a Page.refresh (also see #73). One where you have an XHR response you want to use to partially replace things in the page. And another where you give in a URI which you want to visit / (partially) refresh the current page, with or without some queryParams. 

So 

```coffeescript
Page.refresh
  response: xhr
  onlyKeys: [...]
```

vs

```coffeescript
Page.refresh
  url: ... # optional, defaults to current page
  queryParams: {...} # optional
  onlyKeys: [...] # optional, defaults to full page refresh
```

In the first case we don't do a `Turbolinks.pushState` when the refresh is done, which makes sense because you're not leaving the page. In the bottom case we do update the window state, but only if it is a new URL. 9 out of 10 times that's fine, because either you're visiting a completely new URI, or you're refreshing the current page which doesn't trigger the pushState because the URIs are identical.

### Problem

There is a third case when you are trying to refresh the current page, but with new queryParams. This _does_ get recognized as a new URI, and will do a pushState. For example when you are on https://example.com/admin/article/1 and hit a button that does:

```coffeescript
Page.refresh
  queryParams: {'myAwesomeParam': true}
  onlyKeys: ['a_refresh_block']
```
In this case your browser will update the address field to https://example.com/admin/article/1?myAwesomeParam=true.

I'm arguing that there are situations where you want to refresh the page with a query param without actually touching the window state. For example to toggle a hide/show of something that requires new data from the server. In this PR I introduced a new option on `Page.refresh` that propagates down that prevents the page refresh optionally.  But to be honest I don't think it is common that you want to stay on the same URL, but with different query params, and have that reflected in the browser. In that case we could probably get away with just updating [turbolinks#reflectNewUrl](https://github.com/Shopify/turbograft/blob/78829dd7475ff25e15fdcecee81d466fa0fbcec1/lib/assets/javascripts/turbograft/turbolinks.coffee#L251) to ignore query params and only compare the root URIs.

What do you guys think? @qq99 @nsimmons @celsodantas @richardmonette @DrewMartin Which way should we go?

-----

(A work around is doing the call with $.ajax, and using the resulting XHR in 

```coffeescript
Page.refresh
  response: xhr
  onlyKeys: [...]
```
but I feel turbograft should be able to deal with this)

### DIFF
--- a/lib/assets/javascripts/turbograft/page.coffee
+++ b/lib/assets/javascripts/turbograft/page.coffee
@@ -19,6 +19,7 @@ Page.refresh = (options = {}, callback) ->
   if options.response
     options.partialReplace = true
     options.onLoadFunction = callback
+
     xhr = options.response
     delete options.response
     Turbolinks.loadPage null, xhr, options
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -111,11 +111,12 @@ class window.Turbolinks
 
   @loadPage: (url, xhr, options = {}) ->
     triggerEvent 'page:receive'
+    options.updatePushState ?= true
 
     if doc = processResponse(xhr, options.partialReplace)
-      reflectNewUrl url
+      reflectNewUrl url if options.updatePushState
       nodes = changePage(extractTitleAndBody(doc)..., options)
-      reflectRedirectedUrl(xhr)
+      reflectRedirectedUrl(xhr) if options.updatePushState
       triggerEvent 'page:load', nodes
       options.onLoadFunction?()
     else
--- a/test/javascripts/page_test.coffee
+++ b/test/javascripts/page_test.coffee
@@ -92,7 +92,26 @@ describe 'Page', ->
         response: mockXHR,
         onlyKeys: ['a']
 
-      @replaceStateStub.calledWith mockXHR.getResponseHeader('X-XHR-Redirected-To')
+      assert @replaceStateStub.calledWith sinon.match.any, '', mockXHR.getResponseHeader('X-XHR-Redirected-To')
+
+    it 'doens\'t update window push state if updatePushState is false', ->
+
+      mockXHR = {
+        getResponseHeader: (header) ->
+          if header == 'Content-Type'
+            return "text/html"
+          else if header == 'X-XHR-Redirected-To'
+            return "http://www.test.com/redirect"
+          ""
+        status: 302
+        responseText: "<div>redirected</div>"
+      }
+      Page.refresh
+        response: mockXHR,
+        onlyKeys: ['a']
+        updatePushState: false
+
+      assert @replaceStateStub.notCalled
 
   describe 'onReplace', ->
 
--- a/test/javascripts/turbolinks_test.coffee
+++ b/test/javascripts/turbolinks_test.coffee
@@ -102,6 +102,18 @@ describe 'Turbolinks', ->
         assert @pushStateStub.calledWith({turbolinks: true, url: url_for("/some_request")}, "", url_for("/some_request"))
         assert.equal 0, @replaceStateStub.callCount
 
+      it 'doesn\'t update the window state if updatePushState is false', ->
+        @server.respondWith([200, { "Content-Type": "text/html" }, html_one]);
+
+        Turbolinks.visit "/some_request", {partialReplace: true, onlyKeys: ['turbo-area'], updatePushState: false}
+        @server.respond()
+
+        assert.equal "Hi there!", document.title
+        assert.equal -1, document.body.textContent.indexOf("YOLO")
+        assert document.body.textContent.indexOf("Hi bob") > 0
+        assert @pushStateStub.notCalled
+        assert @replaceStateStub.notCalled
+
       it 'calls a user-supplied callback', ->
         @server.respondWith([200, { "Content-Type": "text/html" }, html_one]);
 


celsodantas:
I think that having an extra param to avoid changing the URL is the way to go. Pagination would be a use case where you want to update the url with just params.

qq99:
I agree with @celsodantas, I like the idea of being able to opt-out of URL updates

maartenvg:
So in that case, what do you guys think of this PR? (I'll have to look into the test failures, and add some specific tests too) Is adding yet another option to the already long list of options something we want to keep doing?

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -248,6 +248,7 @@ class window.Turbolinks
     node
 
   reflectNewUrl = (url) ->
+    debugger
```
nsimmons: oops

lib/assets/javascripts/turbograft/page.coffee
```diff
@@ -19,9 +19,12 @@ Page.refresh = (options = {}, callback) ->
   if options.response
     onlyKeys   = options.onlyKeys   || []
     exceptKeys = options.exceptKeys || []
-    Turbolinks.loadPage null, options.response, true, callback, onlyKeys, exceptKeys
+    Turbolinks.loadPage null, options.response, true, true, callback, onlyKeys, exceptKeys
```
nsimmons: Shouldn't this be false so that we _don't_ update the push state when we have the response already?

maartenvg: It will be ignored later in the stack anyway, but I agree this should be false.

lib/assets/javascripts/turbograft/turbolinks.coffee
maartenvg: oops indeed! Nice catch.

nsimmons:
:+1: to being able to opt out of push state.

lib/assets/javascripts/turbograft/page.coffee
qq99: this method signature is looking sooo messy (not your fault :P)
Would you be able to refactor it to accept an options hash as part of this PR?

qq99: If it is ignored later in the stack, should it be `null` or `undefined` ?

qq99:
I didn't examine the code super closely (navigating down/up the Turbolinks callstack is always a bit of a headache) but it looks solid to me.  Do you want to add something to `remote.coffee` or `initializers.coffee` so that it can work with `tg-remote`, or are you planning to use this as raw `Turbolinks.foo()` type calls?

lib/assets/javascripts/turbograft/page.coffee
nsimmons: I also mentioned the same thing. Options hash is only _slightly_ better though. 

JS needs named params so badly...

maartenvg: I'll look into updating it into a options hash, it is indeed gross. It is not necessarily ignored, but there are other factors that determine whether the window state is updated, and those are never true for XHR-based refreshes.

nsimmons: I say only slightly better because it gets messy dealing with default values for arguments when they are not part of the signature.

maartenvg:
@qq99 The only place I would use this is on explicit Page.refresh calls, but as part of the API it might make sense to add it to the other ways you can end up pushing state.

lib/assets/javascripts/turbograft/page.coffee
qq99: hm true about the defaults :'(  
I guess you can always do something like:
```coffee
foo = (opts) ->
  opts = _.extend({my: 'defaults', in: 'here'}, opts);
```
which is closer to the ruby way

DrewMartin: Side note: coffeescript can kind of fake named params with a hash:
```coffeescript
func = ({foo, bar}) ->
```
which becomes
```javascript
  func = function(_arg) {
    var bar, foo;
    foo = _arg.foo, bar = _arg.bar;
  };
```
Not good for doing default args for them though :/

celsodantas:
I say we leave it for raw `Page.refresh` call only and we can expand it to tg-remote if we need it.
Currently, the use case is only for `Page.refresh` calls

qq99:
:+1:

maartenvg:
I'm cool with that. I'm first going to refactor the arguments hell in a
separate PR, replacing it with an options hash all the way down.

Soleone:
i just ran into this, here's my use case:

basically all i want is a link on a page that does a post request. in normal rails i would just `link_to "Label", url, method: :post` which results in link with `data-method="post"`. as we don't seem to include `jquery_ujs` in shopify (anyone know why by the way?) this doesn't work.

so the workaround seems to be to use `tg-remote="post"` to do the post request. the controller action will end up doing a redirect and it does get followed properly _but_ the location bar is not updated via `pushState` and i'm still seeing the old url where the link was hosted instead of the url of the redirect location.

maartenvg:
Now that #79 is in, this PR becomes extremely trivial. I've also added tests. Please have second review.

nsimmons:
LGTM

qq99:
:ship:

PR #77, opened 2015-03-04
Remove nodes with jQuery if available

DrewMartin:
Because of how we replace nodes anything that has jQuery event handlers attached (explicitly attached, not through Twine) will cause the listening object to leak. For example our Form object does several $.on calls in the constructor and because of this it will never be cleaned up. jQuery event handlers are are only cleaned up when you use jQuery methods to remove the nodes.

When calling [remove](http://api.jquery.com/remove/) and [replaceWith](http://api.jquery.com/replaceWith/) "all bound events and jQuery data associated with the elements are removed"

Review
@Thibaut @qq99 @kristianpd @arthurnn 
We should probably push this upstream to turbolinks. Glancing at their code I believe it will have the same problem
Also I'm not sure how to write tests for memory leaks

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,15 @@ There is an example app that you can boot to play with TurboGraft.  Open the con
 ./server
 ```
 
+## jQuery and memory leaks
+
+When turbograft replaces or removes a node it uses native DOM API to do so. If any objects use jQuery to listen to events on a node then these objects will leak when the node is replaced because jQuery will still have references to it. To clean these up you'll need to tell jQuery that they're removed. This can be done with something like:
+
+```coffeescript
+document.addEventListener 'page:after-node-removed', (event) ->
+  $(event.data).remove()
+```
+
 ## Contributing
 
 1. Fork it ( http://github.com/Shopify/turbograft/fork )
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -48,6 +48,14 @@ if browserSupportsCustomEvents
   installDocumentReadyPageEventTriggers()
   installJqueryAjaxSuccessPageUpdateTrigger()
 
+replaceNode = (newNode, oldNode) ->
+  replacedNode = oldNode.parentNode.replaceChild(newNode, oldNode)
+  triggerEvent('page:after-node-removed', replacedNode)
+
+removeNode = (node) ->
+  removedNode = node.parentNode.removeChild(node)
+  triggerEvent('page:after-node-removed', removedNode)
+
 # TODO: triggerEvent should be accessible to all these guys
 # on some kind of eventbus
 # TODO: clean up everything above me ^
@@ -129,7 +137,7 @@ class window.Turbolinks
         deleteRefreshNeverNodes(body)
 
       triggerEvent 'page:before-replace'
-      document.documentElement.replaceChild body, document.body
+      replaceNode(body, document.body)
       CSRFToken.update csrfToken if csrfToken?
       setAutofocusElement()
       executeScriptTags() if runScripts
@@ -168,7 +176,7 @@ class window.Turbolinks
 
   deleteRefreshNeverNodes = (body) ->
     for node in body.querySelectorAll('[refresh-never]')
-      node.parentNode.removeChild(node)
+      removeNode(node)
 
     return
 
@@ -189,7 +197,7 @@ class window.Turbolinks
 
       if newNode = body.querySelector("##{ nodeId }")
         newNode = newNode.cloneNode(true)
-        existingNode.parentNode.replaceChild(newNode, existingNode)
+        replaceNode(newNode, existingNode)
 
         if newNode.nodeName == 'SCRIPT' && newNode.getAttribute("data-turbolinks-eval") != "false"
           executeScriptTag(newNode)
@@ -197,7 +205,7 @@ class window.Turbolinks
           refreshedNodes.push(newNode)
 
       else if existingNode.getAttribute("refresh-always") == null
-        existingNode.parentNode.removeChild(existingNode)
+        removeNode(existingNode)
 
     refreshedNodes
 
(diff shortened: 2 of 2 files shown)


DrewMartin:
To reproduce in Shopify:
* Visit the product index page
* In the dev tools, open profiles and start recording heap allocations
* Open a particular product
* Save it a few times
* Using the menu bar return to product index
* Stop the heap allocation recording. At the top, enter Form in the Class filter input. You'll see that multiple objects are still in memory

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -129,7 +141,7 @@ class window.Turbolinks
         deleteRefreshNeverNodes(body)
 
       triggerEvent 'page:before-replace'
-      document.documentElement.replaceChild body, document.body
```
qq99: This one is different than all the others; does the function work for this too?

qq99:
This makes sense to me

cc @tylermercier @nsimmons @celsodantas

lib/assets/javascripts/turbograft/turbolinks.coffee
DrewMartin: Yeah. I checked and `document.body.parentNode === document.documentElement`, at least on some pages. I don't know enough about the general dom structure to know if this would always be true though. I could put this a special case if needed

tylermercier:
:+1:

nsimmons:
Did you verify that this fixes the memory leaks?

DrewMartin:
I did, doesn't hurt for someone else to try though

Thibaut:
Sorry but I don't like this. The solution should be to either 1) not attach event listeners to nodes that can be replaced (use live events on the parent), and/or 2) make sure that the objects which attach event listeners get GC-ed at the same time as the DOM nodes (if they only reference each other, with no external reference, they'll get GC-ed together).

That's already the contract Turbolinks sets out at the `<body>` level. One shouldn't hold references to any DOM node outside of the context of that node (e.g. the window). The same contract applies with partial replacement, except at a lower level.

DrewMartin:
Live events?

qq99:
>That's already the contract Turbolinks sets out at the `<body>` level

What do you mean by this?

re: Live events -- they are delegated events bound higher up the DOM that handle events for `n` elements that may or may not exist

Thibaut:
Discussed IRL. The problem is that jQuery holds references to all event listeners it creates, which is forcing leaks on any DOM tree that doesn't get removed by jQuery itself (which does the cleanup).

I guessed this is fine then. I'll try to think of something else but jQuery really isn't helping here…

qq99:
>jQuery holds references to all event listeners it creates
interesting, TIL

DrewMartin:
Yeah, and you have to use one of jQuery's methods to clean them up (html, remove, replaceWith, etc)

tylermercier:
`setTimeout(function(){window.location.reload(1)}, 60000);` :trollface:

pushrax:
If this becomes a thing, the API I'd like to see is exposing a method to set the cleanup function rather than checking for jQuery in a static cleanup function. Lower overhead and cleaner.

Thibaut:
@pushrax I agree. One issue, though, is that I'm not aware of a public jQuery API for cleaning up a tree, other than `$(node).off().find('*').off()`, which sounds inefficient.

I'll bring this up on `rails/turbolinks` this week-end.

qq99:
@pushrax there's probably an event that you could hook into to do your cleanup, but we should guarantee there's exactly 1 reliable event that fires in all the different scenarios

DrewMartin:
Looking at the events that are sent, nothing in there would be great for the cleanup unless you added defers.

If there's a before-replace then you could store the document.body at that time, defer, and then $.remove() it afterwards. Same for the nodes given in partial-replace

What we could do is fire a new event. So keep the methods I added (`replaceNode` and `removeNode`) but instead of having them do a jquery event, fire an event after they replace or remove the node, like `triggerEvent('page:after-remove', replaceNode)`
Then we could listen for that event and clean it up with jQuery

DrewMartin:
Ok updated to fire an event when replacing/removing nodes. What do you guys think? I tested it locally after added this code in to modules.coffee:
```coffeescript
document.addEventListener 'page:after-remove', (event) ->
  removedNode = event.data
  $(removedNode).remove() if removedNode
```
no memory leaks :+1:

qq99:
Just 1 thing with the naming of the event, I think it should be phrased more like `page:after-node-removed`

qq99:
```coffee
document.addEventListener 'page:after-remove', (event) ->
  removedNode = event.data
  $(removedNode).off() if removedNode
```
does this have the same result?  it's a bit more clear that it's about events

DrewMartin:
> does this have the same result? it's a bit more clear that it's about events (try also unbind())

Not quite because it doesn't unbind events on the subtree

DrewMartin:
Updated event name and readme

qq99:
LGTM :+1:

DrewMartin:
@pushrax and @Thibaut cool with using events?

README.md
```diff
@@ -145,6 +145,16 @@ There is an example app that you can boot to play with TurboGraft.  Open the con
...
 ```
 
+## jQuery and memory leaks
+
+When turbograft replaces or removes a node it uses native DOM API to do so. If any objects use jQuery to listen to events on a node then these objects will leak when the node is replaced because jQuery will still have references to it. To clean these up you'll need to tell jQuery that they're removed. This can be done with something like:
+
+```coffeescript
+document.addEventListener 'page:after-node-removed', (event) ->
+  removedNode = event.data
+  $(removedNode).remove() if removedNode
```
Thibaut: why would `removedNode` be `null`?

Thibaut:
Just one question. LGTM :+1: 

Great catch BTW :clap:

README.md
DrewMartin: It shouldn't be, and I guess it wouldn't matter since $(null).remove() is just a no-op

DrewMartin: Updated to be simpler

Thibaut:
:shipit:

(maybe squash)

DrewMartin:
squished

PR #78, opened 2015-03-09
port over Page.onReplace for reliably running a callback after the node in question has been removed from the DOM

qq99:
This allows the user of TurboGraft to register a function to be called *exactly once* when a specific node has been refreshed (be it a full page refresh, or a partial page refresh).

This is mostly a port of what we have in `modules.coffee` in Shopify.  I have not ported `Page.onRefresh` since that definition is more tied to Twine-y things

## Example Usage

```coffee
class MyFoo
  constructor: (@node) ->
    @bindSomeEventsOn(@node)
    Page.onReplace(@node, @unbindSomeEvents.bind(this)) # we want to clean up all the events we bound to prevent leaking

  unbindSomeEvents = ->
    ... # undo whatever we did in the `bindSomeEvents` function
```

cc @DrewMartin for primary review
cc also @patrickdonovan @tylermercier @nsimmons @celsodantas

### DIFF
--- a/lib/assets/javascripts/turbograft/page.coffee
+++ b/lib/assets/javascripts/turbograft/page.coffee
@@ -25,3 +25,47 @@ Page.refresh = (options = {}, callback) ->
 
 Page.open = ->
   window.open(arguments...)
+
+# Providing hooks for objects to set up destructors:
+onReplaceCallbacks = []
+
+# e.g., Page.onReplace(node, unbindListenersFnc)
+# unbindListenersFnc will be called if the node in question is partially replaced
+# or if a full replace occurs.  It will be called only once
+Page.onReplace = (node, callback) ->
+  throw new Error("Page.onReplace: Node and callback must both be specified") if !node || !callback
+  throw new Error("Page.onReplace: Callback must be a function") if !isFunction(callback)
+  onReplaceCallbacks.push({node, callback})
+
+# option C from http://jsperf.com/alternative-isfunction-implementations
+isFunction = (object) ->
+  !!(object && object.constructor && object.call && object.apply)
+
+# roughly based on http://davidwalsh.name/check-parent-node (note, OP is incorrect)
+contains = (parentNode, childNode) ->
+  if parentNode.contains
+    parentNode.contains(childNode)
+  else # old browser compatability
+    !!((parentNode == childNode) || (parentNode.compareDocumentPosition(childNode) & Node.DOCUMENT_POSITION_CONTAINED_BY))
+
+document.addEventListener 'page:before-partial-replace', (event) ->
+  replacedNodes = event.data
+
+  unprocessedOnReplaceCallbacks = []
+  for entry in onReplaceCallbacks
+    fired = false
+    for replacedNode in replacedNodes
+      if contains(replacedNode, entry.node)
+        entry.callback()
+        fired = true
+        break
+
+    unless fired
+      unprocessedOnReplaceCallbacks.push(entry)
+
+  onReplaceCallbacks = unprocessedOnReplaceCallbacks
+
+document.addEventListener 'page:before-replace', (event) ->
+  for entry in onReplaceCallbacks
+    entry.callback()
+  onReplaceCallbacks = []
--- a/test/javascripts/page_test.coffee
+++ b/test/javascripts/page_test.coffee
@@ -85,3 +85,110 @@ describe 'Page', ->
         onlyKeys: ['a']
 
       @replaceStateStub.calledWith mockXHR.getResponseHeader('X-XHR-Redirected-To')
+
+  describe 'onReplace', ->
+
+    beforeEach ->
+      @fakebody = """
+        <div id='fakebody'>
+          <div id='foo'>
+            <div id='bar'>
+              <div id='foobar'></div>
+            </div>
+          </div>
+          <div id='baz'>
+            <div id='bat'></div>
+          </div>
+        </div>
+      """
+      $('body').append(@fakebody)
+
+    afterEach ->
+      $('#fakebody').remove()
+
+
+    it 'calls the onReplace function only once', ->
+      node = $('#foo')[0]
+      refreshing = [$("#fakebody")[0]]
+      callback = stub()
+
+      Page.onReplace(node, callback)
+
+      triggerEvent 'page:before-partial-replace', refreshing
+      triggerEvent 'page:before-replace'
+
+      assert callback.calledOnce
+
+    it 'calls the onReplace function only once, even if someone were to mess with the number of events fired', ->
+      node = $('#foo')[0]
+      refreshing = [$("#fakebody")[0]]
+      callback = stub()
+
+      Page.onReplace(node, callback)
+
+      triggerEvent 'page:before-partial-replace', refreshing
+      triggerEvent 'page:before-partial-replace', refreshing
+      triggerEvent 'page:before-partial-replace', refreshing
+      triggerEvent 'page:before-replace'
+      triggerEvent 'page:before-replace'
+      triggerEvent 'page:before-replace'
+
+      assert callback.calledOnce
+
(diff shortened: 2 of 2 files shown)


lib/assets/javascripts/turbograft/page.coffee
```diff
@@ -25,3 +25,42 @@ Page.refresh = (options = {}, callback) ->
...
   window.open(arguments...)
+
+# Providing hooks for objects to set up destructors:
+onReplaceCallbacks = []
+
+# e.g., Page.onReplace(node, unbindListenersFnc)
+# unbindListenersFnc will be called if the node in question is partially replaced
+# or if a full replace occurs.  It will be called only once
+Page.onReplace = (node, callback) ->
+  throw new Error("Node and callback must both be specified") if !node || !callback
```
DrewMartin: Hmm I wonder if we should also double check that `callback` is a function

lib/assets/javascripts/turbograft/page.coffee
```diff
@@ -25,3 +25,42 @@ Page.refresh = (options = {}, callback) ->
...
+  onReplaceCallbacks.push({node, callback})
+
+# from http://davidwalsh.name/check-parent-node
+contains = (parentNode, childNode) ->
+  if parentNode.contains
+    parentNode.contains(childNode)
+  else
+    parentNode.compareDocumentPosition(childNode) % 16
+
+document.addEventListener 'page:before-partial-replace', (event) ->
```
qq99: this is the block that wasn't straightforward to convert, as we don't have lodash nor jquery here

But I think it may be slightly more performant now with the break

lib/assets/javascripts/turbograft/page.coffee
qq99: http://stackoverflow.com/questions/5999998/how-can-i-check-if-a-javascript-variable-is-function-type/6000016#6000016 ?

lib/assets/javascripts/turbograft/page.coffee
```diff
@@ -25,3 +25,42 @@ Page.refresh = (options = {}, callback) ->
...
+onReplaceCallbacks = []
+
+# e.g., Page.onReplace(node, unbindListenersFnc)
+# unbindListenersFnc will be called if the node in question is partially replaced
+# or if a full replace occurs.  It will be called only once
+Page.onReplace = (node, callback) ->
+  throw new Error("Node and callback must both be specified") if !node || !callback
+  onReplaceCallbacks.push({node, callback})
+
+# from http://davidwalsh.name/check-parent-node
```
DrewMartin: Maybe the comment should also specify that the else is for old firefox compatibility

lib/assets/javascripts/turbograft/page.coffee
```diff
@@ -25,3 +25,42 @@ Page.refresh = (options = {}, callback) ->
...
+  replacedNodes = event.data
+
+  unprocessedOnReplaceCallbacks = []
+  for entry in onReplaceCallbacks
+    fired = false
+    for replacedNode in replacedNodes
+      break if fired
+      if contains(replacedNode, entry.node)
+        entry.callback()
+        fired = true
```
DrewMartin: Let's break after setting fired to true instead of conditionally breaking above

lib/assets/javascripts/turbograft/page.coffee
qq99: Added

test/javascripts/page_test.coffee
```diff
@@ -85,3 +85,93 @@ describe 'Page', ->
...
+            <div id='bat'></div>
+          </div>
+        </div>
+      """
+      $('body').append(@fakebody)
+
+
+    it 'calls the onReplace function only once', ->
+      node = $('#foo')[0]
+      refreshing = $("#fakebody")[0]
```
DrewMartin: Wait does turbograft have a dependancy on jquery? Or only the tests?

DrewMartin: I ask because if so we should let it handle the contains logic above, rather than putting in the browser-specific fixes ourselves

qq99: Only the tests do

lib/assets/javascripts/turbograft/page.coffee
DrewMartin: `typeof func == 'function'` is probably fine
We could go for the hyper performant version but it's not as clear and probably not necessary for this. I don't imagine this would be used thousands of times per second, to the point it was affecting performance, and if it was I would bet the partial replacement node check would have even worse performance.

lib/assets/javascripts/turbograft/page.coffee
```diff
@@ -25,3 +25,47 @@ Page.refresh = (options = {}, callback) ->
...
+# option C from http://jsperf.com/alternative-isfunction-implementations
+isFunction = (object) ->
+  !!(object && object.constructor && object.call && object.apply)
+
+# from http://davidwalsh.name/check-parent-node
+contains = (parentNode, childNode) ->
+  if parentNode.contains
+    parentNode.contains(childNode)
+  else # old browser compatability
+    parentNode.compareDocumentPosition(childNode) % 16
```
DrewMartin: I'm skeptical that this works. I say that because I tried it and it doesn't

qq99: Fixed via comments on that article:

```
parent.compareDocumentPosition(child) & 16; // 16, truthy
child.compareDocumentPosition(parent) & 16; // 0, falsy
```

DrewMartin:
:+1: 
Side note: it does seem kind of strange that turbograft is listening to its own events and reacting. We could just as well handle it inside turbograft/turbolinks.coffee

DrewMartin:
Nice tests

lib/assets/javascripts/turbograft/page.coffee
```diff
@@ -25,3 +25,47 @@ Page.refresh = (options = {}, callback) ->
...
+# option C from http://jsperf.com/alternative-isfunction-implementations
+isFunction = (object) ->
+  !!(object && object.constructor && object.call && object.apply)
+
+# roughly based on http://davidwalsh.name/check-parent-node (note, OP is incorrect)
+contains = (parentNode, childNode) ->
+  if parentNode.contains
+    parentNode.contains(childNode)
+  else # old browser compatability
+    !!((parentNode == childNode) || (parentNode.compareDocumentPosition(childNode) & 16))
```
DrewMartin: We can change the 16 to `Node.DOCUMENT_POSITION_CONTAINED_BY`
https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition

qq99:
:shipit: ?

test/javascripts/page_test.coffee
```diff
@@ -85,3 +85,110 @@ describe 'Page', ->
...
+      triggerEvent 'page:before-partial-replace', refreshing
+      triggerEvent 'page:before-partial-replace', refreshing
+      triggerEvent 'page:before-replace'
+      triggerEvent 'page:before-replace'
+      triggerEvent 'page:before-replace'
+
+      assert callback.calledOnce
+
+    it 'throws an error not supplied enough arguments', ->
+      try
```
DrewMartin: Are we using chai here for assertions? It has assert throw functionality: http://chaijs.com/api/assert/#throws-section

qq99: I think it's a lot more straightforward to test the errors this way (test em as you'd use em)

Is that alright?  Not sure that I could test the string message without a lot of syntax

DrewMartin: oh true

qq99:
so, reping for :ship:?  Not sure what's up with the travis CI hooks now (somehow, 2 lines now appear), but it looks like all is well

DrewMartin:
:ship:

PR #79, opened 2015-03-11
Replace long argument list with options hash for fetch/loadPage/etc

maartenvg:
If we want to add more options to to Page.refresh, the number of arguments on Turbograf#loadPage/Turbograft#fetch and it's descendants will grow into an unmaintainable mess. So when doing https://github.com/Shopify/turbograft/pull/74 it was suggested we introduce option hashes throughout. 

This PR does that, to keep https://github.com/Shopify/turbograft/pull/74 clean of unrelated changes. I tried to reuse the same options hash all the way down.

Please review @qq99 @nsimmons @DrewMartin


nsimmons:
Much more readable!

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -106,33 +109,33 @@ class window.Turbolinks
 
     return
 
-  @loadPage: (url, xhr, partialReplace = false, onLoadFunction = (->), replaceContents = [], replaceAllExcept = []) ->
+  @loadPage: (url, xhr, options = {}) ->
```
qq99: should url and xhr go into options hash here?
mainly because I noticed above `url` is passed in as `null`

qq99:
:heart:

lib/assets/javascripts/turbograft/turbolinks.coffee
maartenvg: I find they are less options, and more of the core signature of this method. So I prefer to leave it like this.

qq99:
There were some test failures, fixed em up :)

test/javascripts/remote_test.coffee
```diff
@@ -91,6 +91,7 @@ describe 'Remote', ->
       assert.equal "anything", request.requestHeaders["X-Header"]
 
     it 'will automatically set the X-CSRF-Token header for you', ->
```
qq99: interestingly, this test failed on the Firefox that I have inside a VM

maartenvg:
I accidentally force pushed over your commit. Hadn't seen your commits yet (we just happened to work on it at the same time). If you want you can force push your last changes.

qq99:
done :)

qq99:
:clap: thanks!

PR #80, opened 2015-03-12
add tg-remote-noserialize attribute

richardmonette:
Currently, to avoid serialization, an input can either a) not have a name attribute or b) have the disabled attribute. This is problematic in situations where an input should have a name and be enabled, such that it can be edited, but may not need to serialized. 

This PR introduces `tg-remote-noserialize` which, when it appears on an input, will cause it to be omitted from serialization.

@qq99 @pushrax @nsimmons @DrewMartin @maartenvg

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,19 @@ Examples of where this may be useful include:
 
 For the lazy developer in all of us, we can use the attribute `refresh-always` when we want to be sure we've absolutely replaced a certain element, if it exists.  An example of such a node you may want to apply this might be an unread notification count -- always being sure to update it if it exists in the response.
 
+### tg-remote-noserialize
+
+When serializing forms for tg-remote calls, turbograft will check to ensure inputs meet the following criteria:
+
+- the input has a `name` attribute
+- the input does not have the `disabled` attribute
+
+and
+
+- the input does not have the `tg-remote-noserialize`
+
+The `tg-remote-noserialize` is useful in scenarios where the input should be editable, i.e. not `disabled`, but should also not be submitted to the server.
+
 ## Example App
 
 There is an example app that you can boot to play with TurboGraft.  Open the console and network inspector and see it in action!  This same app is also used in the TurboGraft browser testing suite.
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -92,7 +92,7 @@ class TurboGraft.Remote
   _iterateOverFormInputs: (form, callback) ->
     inputs = form.querySelectorAll("input:not([type='reset']):not([type='button']):not([type='submit']):not([type='image']), select, textarea")
     for input in inputs
-      inputEnabled = !input.disabled
+      inputEnabled = !input.disabled && !input.hasAttribute('tg-remote-noserialize')
       radioOrCheck = (input.type == 'checkbox' || input.type == 'radio')
 
       if inputEnabled && input.name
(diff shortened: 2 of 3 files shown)


DrewMartin:
Test it

qq99:
When would it need a name but not need to be serialized?

richardmonette:
@qq99 I'm writing all this stuff for https://github.com/Shopify/shopify/pull/40387/files

I suppose we could technically be adding and removing the name attribute, whereas right now I was planning to use `tg-no-serialize`. It seemed cleaner that the name attribute would always exist and there would be a particular attribute which controls serialization.

DrewMartin:
We basically want to conditionally serialize an input only when some other inputs have a value

qq99:
Yeah, I think we discussed a `tg-remote-noserialize` in the past, I was in favour of it but others said you should just disable the `input`

I'm cool with the idea for extra explicitness

qq99:
cc @celsodantas

qq99:
Here was more discussion around `tg-remote-noserialize` for cross-reference purposes

qq99:
Can you add a note about this to the README please?

celsodantas:
`disable` have a visual impact on the HTML element (the browser will usually grey out the input, if no CSS to avoid that), so I'm in favor of a `tg-no-serialize` too.

richardmonette:
@celsodantas exactly, we need a disable which has no impact on the browser visualization

richardmonette:
@DrewMartin added tests
@qq99 updated the README

qq99:
Perfect :sparkles: :ship:

richardmonette:
@nsimmons You and @DrewMartin had some objections to this in https://github.com/Shopify/turbograft/pull/66 however we've got another case where this could be useful. I believe @DrewMartin supports the idea now - just wanted to make sure you are :cool: with this too.

nsimmons:
I hate that we have to add this attribute, but i understand the reasons. 

Why are we against using CSS to prevent the visual difference with disabled inputs?

maartenvg:
It's not just visual. In some cases they should still be able to fill out data in the field we would mark with this.

nsimmons:
We want to be able to fill in data that won't be submitted to the server?

DrewMartin:
shop one products new shopify 2015-03-13 12-58-29

Here's the option/value entry for new variants. If the user has entered no option values we don't want to serialize the option names, instead we want to just have one, default variant since that shows a cleaner view. Also if we serialize the empty option values we get validation errors and it's confusing for the merchant. We've been getting feedback where they click "add variants" and then don't enter anything and the product fails to save. We should allow it, and treat it as if they didn't click add variants

nsimmons:
I see.

This sounds like something that would not be that common of a case though. Do you think this is something we should be building into Turbograft, or could we just handle this specific detail on this page with some custom JS?

qq99:
I think it could be handled another way, but the code and complexity impact of this is small enough that I don't see a reason not to add it to TurboGraft

nsimmons:
It is small. It just feels like an edge case and not something that is core to Turbograft.

I don't want to block this, and the change is additive so if you don't know about this option it won't break anything. It does feel like a slippery slope though.

DrewMartin:
Yeah, it wouldn't be that useful in many places right now. We do want to make it able to disable serialization for a whole subtree. That would actually be super useful as long as we can make it performant enough.

As for why: Consider a place where you have a whole bunch of inputs that are all disabled on the same condition. Being able to disable an element that contains them all would be safer than disabling them individually because if more inputs are added the disabled condition doesn't need to be added to it, as well.

richardmonette:
I'm using this in https://github.com/Shopify/shopify/pull/40387 where I think I can work around without needing this feature.

maartenvg:
I think enough has been said that indicates that at this point we don't need nor want this. Closing.

PR #81, opened 2015-03-17
open up a hook for other plugins to define their own replaceNode functionality

qq99:
[Turbo-React](https://github.com/ssorallen/turbo-react) is a small repo (https://github.com/ssorallen/turbo-react/tree/master/src) with 2 main needs (see also https://github.com/ssorallen/turbo-react/issues/19).  Check out this demo!  https://turbo-react.herokuapp.com/

- [ ] Turbolinks should expose `Turbolinks.replaceNode(...)` as the public API, so a monkey patch of the DOM isn't needed here https://github.com/ssorallen/turbo-react/blob/e0048b78d2c915d0517efff1afdf9c6cf7b6ca75/src/reactize.js#L54
- [ ] https://github.com/ssorallen/turbo-react/blob/e0048b78d2c915d0517efff1afdf9c6cf7b6ca75/src/reactize.js#L24-L30 should be changed to call the original `Turbolinks.replaceNode(...)`

I think it'd be nice to provide this hook

### DIFF
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -48,10 +48,6 @@ if browserSupportsCustomEvents
   installDocumentReadyPageEventTriggers()
   installJqueryAjaxSuccessPageUpdateTrigger()
 
-replaceNode = (newNode, oldNode) ->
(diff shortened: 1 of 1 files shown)


celsodantas:
:+1: 

Should we add a test to make sure that the method is a class method?

qq99:
We can consider this delayed for now until more work happens on turbo-react

PR #82, opened 2015-03-20
Add attribute that prevents serialization of a subtree

DrewMartin:
Modified version of #80

This adds a `tg-remote-noserialize` attribute, but unlike the last one it prevents serialization of a node and its subtree. Consider a field that has many inputs inside of it which you want to serialize together, conditionally. Either they're all sent or none of them are. Rather than putting disabled on them individually we can just put `tg-remote-noserialize` on some ancestor of them.

My biggest concern for this was performance. I tried a few different methods and benchmarked them [here](http://jsperf.com/finding-enabled-inputs-by-attribute/3). Tl;dr this method's the fastest in general. In practical terms I tried making a product with 100 variants and 100+ tags, for about 1300 inputs within the main product#show form. I put `tg-remote-noserialize` on each variant row individually, so 100 instances of it on the page. Getting the enabled inputs took between 8-15 ms per run. With only two rows with the attribute it was 2-4 ms.

Performance could be improved further if we were to use Set or WeakSet (since any object can be a key and uniquely identified) but it's still experimental and doesn't have widespread support. Assuming `n` is the number of inputs and `m` is the number of nodes with disabled attributes, it could take the run time from `O(m*n)` to `O(m+n)`. With the speeds we're seeing it would probably be unnecessary though.

TODO:
* [ ] Docs

Review: @qq99 @richardmonette 
@Shopify/admin

### DIFF
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -90,7 +90,7 @@ class TurboGraft.Remote
     formData
 
   _iterateOverFormInputs: (form, callback) ->
-    inputs = form.querySelectorAll("input:not([type='reset']):not([type='button']):not([type='submit']):not([type='image']), select, textarea")
+    inputs = @_enabledInputs(form)
     for input in inputs
       inputEnabled = !input.disabled
       radioOrCheck = (input.type == 'checkbox' || input.type == 'radio')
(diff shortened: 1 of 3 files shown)


lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -99,6 +99,22 @@ class TurboGraft.Remote
         if (radioOrCheck && input.checked) || !radioOrCheck
           callback(input)
 
+  _enabledInputs: (form) ->
+    selector = "input:not([type='reset']):not([type='button']):not([type='submit']):not([type='image']), select, textarea"
+    inputs = Array.prototype.slice.call(form.querySelectorAll(selector))
```
DrewMartin: Doing this `Array.prototype.slice.call` stuff because form.querySelectorAll does not return an array, but a NodeList. NodeLists lacks some features of arrays like the ability to modify them with concat. This convert it to an array.

lemonmade: `<something>.prototype` can be condensed to `<something>::` in CS, so you can write this a little more succinctly with `Array::slice.call(form.querySelectorAll(selector))`

qq99:
LGTM :+1:

richardmonette:
:cool: :+1:

lib/assets/javascripts/turbograft/remote.coffee
DrewMartin: Updated

DrewMartin:
Updated the readme

qq99:
:+1:

nsimmons:
Still have the same concerns I did from #80. Not convinced this should be part of turbograft.

DrewMartin:
Unless we pull out TurboGraft.Remote of TurboGraft to get this functionality we'll either need to do it here, or do a pretty ugly monkey patch that changes the `_iterateOverFormInputs` method inside from inside shopify

Thibaut:
> Unless we pull out TurboGraft.Remote of TurboGraft

That would be a great move. We'll need to do this when we switch back to Turbolinks 3.

qq99:
So... :ship: ? lol

DrewMartin:
> That would be a great move. We'll need to do this when we switch back to Turbolinks 3.

Much bigger change. If it's in here then if/when we do that it'll be part of the copied code :D

PR #83, opened 2015-04-02
Before replace node event

tmlayton:
See Shopify/shopify#41948

@qq99 @nsimmons

### DIFF
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -49,7 +49,7 @@ if browserSupportsCustomEvents
   installJqueryAjaxSuccessPageUpdateTrigger()
 
 replaceNode = (newNode, oldNode) ->
-  triggerEvent('page:before-node-removed', {newNode: newNode, oldNode: oldNode}
+  triggerEvent('page:before-node-removed', {newNode: newNode, oldNode: oldNode})
   replacedNode = oldNode.parentNode.replaceChild(newNode, oldNode)
   triggerEvent('page:after-node-removed', replacedNode)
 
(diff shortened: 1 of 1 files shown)


qq99:
:+1:

maartenvg:
:+1:

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -49,6 +49,7 @@ if browserSupportsCustomEvents
   installJqueryAjaxSuccessPageUpdateTrigger()
 
 replaceNode = (newNode, oldNode) ->
+  triggerEvent('page:before-node-removed', {newNode: newNode, oldNode: oldNode})
```
DrewMartin: We should probably fire this event inside removeNode as well, considering the name. Otherwise the event should be renamed to `page:before-node-replaced`

qq99: :+1:

tmlayton: updated

tmlayton:
Okay to :ship:? And how do I go about making a release?

qq99:
https://github.com/Shopify/turbograft/blob/master/release.md after you merge

maartenvg:
What about tests?

jduff:
@Thibaut are we going to need to add this to turbolinks? Should we do this somewhere else instead right  now so that we don't diverge more?

DrewMartin:
Yeah, we'll need it in turbo links eventually

Thibaut:
@jduff `after-remove` is in already: https://github.com/rails/turbolinks/pull/485
I can PR `before-remove` over the week-end but I want to understand the use-case first. See https://github.com/Shopify/shopify/pull/41948#issuecomment-89027357

tmlayton:
Thanks to @Thibaut's suggestion found a way to sort out Shopify/shopify#41948 without this.

PR #85, opened 2015-04-15
add some strategems

qq99:
Nicely rendered view: https://github.com/Shopify/turbograft/blob/tg-docs/docs/strategems.md

### DIFF
--- a/docs/strategems.md
+++ b/docs/strategems.md
@@ -1,0 +1,273 @@
+# TurboGraft Stratagems
+
+## Strategems through example
+
+### Remote partials
+
+*Potential Desires:*
+- Avoid rendering an expensive partial inline, fetch from the server on click
+- `Show more...` type links
+
+*Example:*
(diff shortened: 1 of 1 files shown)


qq99:
cc @Shopify/tnt thoughts? comments? criticisms?

At some point I thought this might be useful, but now I'm less satisfied with it now that I've written it.  But I've gone past the point of no return so here's a PR!

Any things I should add?  I know it's missing `Perform an action and partially refresh` and an example of refreshing multiple nodes using a single refresh key or refreshing multiple nodes using multiple refresh keys, but, I'm not sure that's necessary?

celsodantas:
I like the idea. But maybe we should keep it in the wiki?

qq99:
Great point lol

qq99:
https://github.com/Shopify/turbograft/wiki/TurboGraft-Strategems

PR #88, opened 2015-04-22
add option to specify using native encoding

richardmonette:
There are certain situations where it is necessary to use native FormData encoding where `Turbograft.Remote` would not have automatically selected that type of form encoding (https://github.com/Shopify/shopify/pull/43275). 

Therefore, this PR adds an option to `Turbograft.Remote` which allows for specifying that native encoding being used. When not specified, the behaviour is the same as before, i.e. attempt to use URL encoding except when input fields are present.

review @qq99 @DrewMartin @maartenvg 
cc @Shopify/tnt

### DIFF
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -4,6 +4,7 @@ class TurboGraft.Remote
     @initiator = form || target
 
     @actualRequestType = if @opts.httpRequestType?.toLowerCase() == 'get' then 'GET' else 'POST'
+    @useNativeEncoding = @opts.useNativeEncoding
 
     @formData = @createPayload(form)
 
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -4,6 +4,7 @@ class TurboGraft.Remote
     @initiator = form || target
 
     @actualRequestType = if @opts.httpRequestType?.toLowerCase() == 'get' then 'GET' else 'POST'
+    @useNativeEncoding = @opts.useNativeEncoding if @opts.useNativeEncoding
```
pushrax: Don't need the `if` here

pushrax:
:+1:

maartenvg:
:+1:

qq99:
:+1: 
Your form had no `file` in it?  Can you explain more about the use case?

richardmonette:
For collections images, when the image is dragged and dropped, I am catching that event and storing the file. Typically, that file would then be sent with an Ajax request to the server. However, I'd like to send the file along with the form in this case. By making my own Turbograft.Remote instance, and adding the file to the FormData, this can be achieved. The one trick is that I need to make sure the FormData is natively encoded, such that I can attach my file.

qq99:
Gotcha https://github.com/Shopify/shopify/pull/43275/files#diff-7a10bd70c27fa5030a0e75641b1ac4d7R21

PR #89, opened 2015-05-27
Update README.md

johnrees:
suggest removing turbolinks references from a rails project as it can cause systemstack errors


kurtfunai:
:+1:

pushrax:
Thanks!

PR #90, opened 2015-05-27
Fix build and deprecation warnings

pushrax:
Teaspoon [updated their package structure](https://github.com/modeset/teaspoon/blob/master/CHANGELOG.md#100-5415). I also fixed a couple Rails deprecation warnings.

@qq99


qq99:
:+1:

PR #92, opened 2015-05-29
make turbolinks run scripts of child nodes on replace

kristianpd:
WIP

This makes sure any scripts nested in a node being replaced are run unless they're not supposed to.


pushrax:
Rebasing off master will fix the build.

akrav:
@kristianpd Can you get this change into master?

qq99:
Looks legit :+1:

akrav:
cc @Shopify/tnt
@pushrax Can you get this change into master?

patrickdonovan:
The test failures are legit - just guessing but it's probably because in the case when the root node of the swapped-in html is a script tag, it gets missed by `node.querySelectorAll 'script:not([data-turbolinks-eval="false"])'` Haven't had time to fix it.

PR #93, opened 2015-05-29
Safari form parsing fix from turbolinks

patrickdonovan:
@qq99 @pushrax @kristianpd 

I'm running into [this issue from turbolinks](https://github.com/rails/turbolinks/issues/408) on a page in shopify. This PR snipes [this fix](https://github.com/rails/turbolinks/commit/faa962329ec2b18817034db01bebe01ee67f3333) for the issue.

### DIFF
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -242,6 +242,15 @@ class window.Turbolinks
     keepNodes(body, allNodesToKeep)
     return
 
+  createDocumentUsingFragment = (html) ->
+    head = html.match(/<head[^>]*>([\s\S.]*)<\/head>/i)?[0] or '<head></head>'
+    body = html.match(/<body[^>]*>([\s\S.]*)<\/body>/i)?[0] or '<body></body>'
+    htmlWrapper = document.createElement 'html'
+    htmlWrapper.innerHTML = head + body
+    doc = document.createDocumentFragment()
+    doc.appendChild htmlWrapper
+    doc
+
   executeScriptTags = ->
     scripts = Array::slice.call document.body.querySelectorAll 'script:not([data-turbolinks-eval="false"])'
     for script in scripts when script.type in ['', 'text/javascript']
@@ -303,7 +312,7 @@ class window.Turbolinks
       xhr.getResponseHeader('Content-Type').match /^(?:text\/html|application\/xhtml\+xml|application\/xml)(?:;|$)/
 
     extractTrackAssets = (doc) ->
-      for node in doc.head.childNodes when node.getAttribute?('data-turbolinks-track')?
+      for node in doc.querySelector('head').childNodes when node.getAttribute?('data-turbolinks-track')?
         node.getAttribute('src') or node.getAttribute('href')
 
     assetsChanged = (doc) ->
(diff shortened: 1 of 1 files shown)


qq99:
oh nice this might also fix `<tr>` coming back

Thibaut:
The parsing code in Turbolinks has since then become much simpler:
https://github.com/rails/turbolinks/blob/a2b1810a60fea0ea131eea39fae65a771f2938a9/lib/assets/javascripts/turbolinks.js.coffee#L330-L339

kristianpd:
^ grab the latest version @Thibaut linked

patrickdonovan:
OK, I pulled the latest version of `createDocument` from turbolinks. It replaces the entirety of `browserCompatibleDocumentParser`. My own :tophat:-testing passes for our page in all browsers using this branch in shopify.

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -335,48 +335,21 @@ class window.Turbolinks
...
-    finally
-      unless testDoc?.body?.childNodes.length is 1
-        return createDocumentUsingWrite
+  createDocument = (html) ->
+    if /<(html|body)/i.test(html)
+      doc = document.documentElement.cloneNode()
+      doc.innerHTML = html
+    else
+      doc = document.documentElement.cloneNode(true)
+      doc.querySelector('body').innerHTML = html
```
celsodantas: I might be without context here, but I don't know (by just looking at this code) what is the difference between the code in the `if` and the `else`. Maybe move them to a function with a descriptive name a new dev can understand the difference? (or maybe just comments)

patrickdonovan: This is taken straight out of turbolinks master so I think there's merit to leaving it as-is - or adding comments at most. That said I'm not 100% sure either but it looks like the else half does a deep clone of the document's root element because `<body` or `<html` are not present in the argument html, and sets the argument html to be a child of the copied current document's `<body>`, whereas the if half just clones the root node and sets the passed-in html as the child of that.

celsodantas: blame rails for the encrypted code then =P

DrewMartin:
Does this mean no more of those `createDocument is not a function` errors? :heart_eyes:

lib/assets/javascripts/turbograft/turbolinks.coffee
Thibaut: In one case the argument is an HTML _document_ string, in the other case it's an HTML _fragment_ string. The second case was added when partial replacement was introduced (since the server doesn't render the full page all the time anymore).

celsodantas: hard to understand if you don't have that context =/

Thibaut: If find `if /<(html|body)/i.test(html)` quite understandable :P

celsodantas: you are too smart.

patrickdonovan:
> Does this mean no more of those createDocument is not a function errors? :heart_eyes:

Seems likely.

So uh... :ship: ? We'll do a thorough :tophat: on the version bump PR in shopify.

pushrax:
:+1:

Issue #96, opened 2015-07-15
`rememberReferer` does not always seem to be performing correctly


qq99:
If you're using a `tg-remote`, I believe it will always call `Turbolinks.pushState` (https://github.com/Shopify/turbograft/blob/master/lib/assets/javascripts/turbograft/turbolinks.coffee#L271)

That is to say, the check for `current url !== referer` will always pass, in most cases.

1.  Perform a regular navigation from A to B
2.  If you break in `reflectNewUrl`, you can see that `referer` was A, and the new page is B (this is the good case, all as expected)
3.  On page B, you click a button that looks like this:
```html
<a tg-remote="POST" href="/resource/1/some_action">do something and refresh this page</a>
```
For this, `some_action` renders `show` but doesn't redirect you anywhere.
4. In your break point, since we haven't performed a regular navigation, `referer` hasn't been updated and is still A.
5. The exact same URL, B, is pushed onto the history stack, and the back button brings you to the same page you were just on

I believe that `tg-remote` should call `rememberReferer` (https://github.com/Shopify/turbograft/blob/master/lib/assets/javascripts/turbograft/turbolinks.coffee#L281-L282) at some point.  Perhaps before it makes the XHR, or perhaps after the `success` of the XHR but before calling `Page.refresh`?

`rememberReferer` isn't currently a public method

cc @Shopify/tnt 

Investigating this, it reminded me of https://github.com/Shopify/turbograft/pull/74 which is similar insofar as they both deal with history.  The easy no-conflict solution would be to provide an attribute that let's you do `tg-no-historystate` or something and leverage the code from #74, but I wonder if we can't just `rememberReferer` with no problems?

Thoughts?  Any potential side effects?

qq99:
Actually, just realized maybe I can add a browser test instead of a unit test

TheMallen:
This was fixed in #97, closing.

PR #97, opened 2015-07-16
Fix historyState bug with tg-remote

qq99:
Fixes https://github.com/Shopify/turbograft/issues/96

If you were to call `Turbolinks.loadPage` directly (bypassing `Turbolinks.visit`), `rememberReferer` would not be called.  `tg-remote` does exactly this via `Page.refresh` and supplying an `xhr` in `page.coffee`.

This PR duplicates a call to `rememberReferer`, but should have no effect in the `Turbolinks.visit` case since it's just performing the same thing twice.  It's feels a little unfortunate that it had to be duplicated

I think that this is a proper fix because:

- if you navigate from page A > B > C without this patch, a `tg-remote` that didn't result in a redirect would have `B` as the referer, and re-push `C` onto the stack to result in A > B > C > C
- a partial replace should probably care about its referer much in the same way that a regular navigation cares about its referer

I've tested it out on my own project with the back & forward buttons in these scenarios:

- regular navigations
- tg-remote with only an action (aka `xhr` but no `onlyKeys`)
- tg-remote with an action and `refresh-on-success` (`xhr` + `onlyKeys`)

Also good for testing is the example app (added a new button + test for this kind of behaviour)

*Don't merge*: still need to fix:

- tg-remote GET where you want to replace parts of the page and update the URL in history
- tg-remote GET where you want to replace parts of the page and keep URL as it was (no history altering)

cc @Shopify/tnt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,4 @@ env:
   - DISPLAY=':99.0'  RAILS_VERSION='4.1.0'
   - DISPLAY=':99.0'  RAILS_VERSION='4.2.0'
 script: bundle exec rake ci
+sudo: false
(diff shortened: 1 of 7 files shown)


qq99:
Actually, just realized maybe I can add a browser test instead of a unit test

qq99:
Closing this, too old, forgot what I was doing with it exactly.
Lives on in https://github.com/Shopify/turbograft/pull/156

Issue #101, opened 2015-09-10
Script included in refresh nodes is not executed


akrav:
If you include script tag in nodes that needs to be refreshed, it isn't executed.

PR #102, opened 2015-09-10
Script execution included in nodes refresh.

akrav:
Issue: https://github.com/Shopify/turbograft/issues/101


patrickdonovan:
This would run all the scripts on the page wouldn't it? We ran into this once a while ago and had https://github.com/Shopify/turbograft/pull/92 as a WIP, which would run the scripts only in the node you're swapping in.

akrav:
Closing this PR since https://github.com/Shopify/turbograft/pull/92 fixes it.
Thanks @patrickdonovan

PR #104, opened 2015-10-22
Don’t set request_method cookie for GET requests

wvanbergen:
The presence of a `Set-Cookie` header prevents CDNs from caching a page, so we want to prevent cookies from being set unnecessarily. Turbolinks/Turbograft will assume a request is a GET request if the cookie is not set, so doesn’t doesn’t change Turbograft’s behaviour. 

This mirrors https://github.com/rails/turbolinks/commit/62cc3db457ad9745ccda047dd43ab84fb3d89707

Review: @pushrax @patrickdonovan, cc @mtaher @Shopify/cloud


Thibaut:
:+1:

patrickdonovan:
:+1:

PR #108, opened 2015-11-16
support data attrs

### LINKED ISSUE #105: Allow `data-` attrs for valid HTML5
In the spirit of https://github.com/Shopify/twine/issues/43. To paraphrase:

> No page that uses Turbograft is valid HTML because of the custom attributes. I'm not suggesting killing these unprefixed versions, but it would be great if we could alias all of Turbograft’s attributes (e.g. refresh) to data attributes (e.g. data-refresh). That way any users that care about having valid markup can use the data- ones, and those that don’t can continue using the unprefixed versions.

nwtn:
Fixes #105 by supporting `data-*` for all the TG attributes

This is actually a fairly inelegant commit, and I'm definitely open to suggestions about how to do this more cleanly. I haven't bothered to write new tests yet because I wanted to get a read on what others thought of this approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,19 +29,19 @@ Turbograft was built with simplicity in mind. It intends to offer the smallest a
 ### Partial page refresh
 
 ```html
-<div id="content" data-refresh="page">
+<div id="content" data-tg-refresh="page">
   ...
 </div>
 ```
 
 
 ```html
-<a href="#" id="partial-refresh-page" data-refresh="page" onclick="event.preventDefault(); Page.refresh({url: '<%= page_path(@next_id) %>',onlyKeys: ['page']});">Refresh the page</a>
+<a href="#" id="partial-refresh-page" data-tg-refresh="page" onclick="event.preventDefault(); Page.refresh({url: '<%= page_path(@next_id) %>',onlyKeys: ['page']});">Refresh the page</a>
 ```
 
-This performs a `GET` request, but our client state is maintained. Using the `data-refresh` attribute, we tell TurboGraft to grab the new page, but only refresh elements where `data-refresh="page"`.  This is the lowest-level way to use TurboGraft.
+This performs a `GET` request, but our client state is maintained. Using the `data-tg-refresh` attribute, we tell TurboGraft to grab the new page, but only refresh elements where `data-tg-refresh="page"`.  This is the lowest-level way to use TurboGraft.
 
-`data-refresh` attributes on your DOM nodes can be considered somewhat analoguous to how `class` will apply styles to any nodes with that class.  That is to say, many nodes can be decorated `data-refresh="foo"` and all matching nodes will be replaced with `onlyKeys: ['foo']`.  Each node with `data-refresh` must have its own unique ID (this is how nodes are matched during the replacement stage).  At the moment, `data-refresh` does not support multiple keys (e.g., `data-refresh="foo bar"`) like the `class` attribute does.
+`data-tg-refresh` attributes on your DOM nodes can be considered somewhat analoguous to how `class` will apply styles to any nodes with that class.  That is to say, many nodes can be decorated `data-tg-refresh="foo"` and all matching nodes will be replaced with `onlyKeys: ['foo']`.  Each node with `data-tg-refresh` must have its own unique ID (this is how nodes are matched during the replacement stage).  At the moment, `data-tg-refresh` does not support multiple keys (e.g., `data-tg-refresh="foo bar"`) like the `class` attribute does.
 
 ### onlyKeys
 You can specify multiple refresh keys on a page, and you can tell TurboGraft to refresh on one or more refresh keys for a given action.
@@ -57,12 +57,12 @@ You can also tell TurboGraft to refresh the page, but exclude certain elements f
 <button id='refresh-a-and-b' href="<%= page_path(@id) %>" onclick="event.preventDefault(); Page.refresh({url: '<%= page_path(@id) %>', exceptKeys: ['section-a', 'section-b']});">Refresh everything but Section A and B</button>
 ```
 
-### data-refresh-never
-The `data-refresh-never` attribute will cause a node only appear once in the `body` of the document. This can be used to include and initialize a tracking pixel or script just once inside the body.
+### data-tg-refresh-never
+The `data-tg-refresh-never` attribute will cause a node only appear once in the `body` of the document. This can be used to include and initialize a tracking pixel or script just once inside the body.
 
 ```html
-<div data-refresh-never>
-  <%= link_to "Never refresh", page_path(@next_id), id: "next-page-refresh-never", "data-refresh" => "page" %>
+<div data-tg-refresh-never>
+  <%= link_to "Never refresh", page_path(@next_id), id: "next-page-refresh-never", "data-tg-refresh" => "page" %>
 </div>
 ```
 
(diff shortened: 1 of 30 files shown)


nwtn:
test results seem inconsistent and i can't reproduce locally. any help understanding what's happening with this test would be awesome.

nwtn:
woo tests passing

nwtn:
ping @Shopify/tnt for review

README.md
```diff
@@ -57,54 +57,54 @@ You can also tell TurboGraft to refresh the page, but exclude certain elements f
...
+* `data-full-refresh-on-success-except`: (optional) Replaces body except for specififed refresh keys, using the body of the XHR which has succeeded
+* `data-refresh-on-error`: (optional) The refresh keys to be refreshed, but using body of XHR which has failed. Only works with error 422. If the XHR returns and error and you do not supply a refresh-on-error, nothing is changed
+* `data-full-refresh-on-error-except`: (optional) Replaces body except for specified refresh keys, using the body of the XHR which has failed.  Only works with error 422
+* `data-remote-once`: (optional) The action will only be performed once. Removes `data-remote-method` and `data-remote-once` from element after consumption
+* `data-full-refresh`: Rather than using the content of the XHR response for partial page replacement, a full page refresh is performed. If `data-refresh-on-success` is defined, the page will be reloaded on these keys. If `data-refresh-on-success` is not defined, a full page refresh is performed. Defaults to true if neither refresh-on-success nor refresh-on-error are provided
```
qq99: It's unfortuante that, for whatever reason when I implemented this, I didn't scope them all `tg-*`... :cry:

qq99:
The code looks failproof to me, just `||` conditions

The only thing I'd suggest is that while we're doing it, use `data-tg-*` for everything

nwtn:
Namespaced everything. If this approach seems OK I'll do some tests and re-ping shortly.

nwtn:
switched back to `getAttribute`, updated tests. @Shopify/tnt

qq99:
:+1: from me

DrewMartin:
Nothing in the docs suggests that the old attribute names are still valid so this looks like a breaking change just by the docs. There should probably be something in there about the old names (or make them deprecated?)

lib/assets/javascripts/turbograft/initializers.coffee
```diff
@@ -4,34 +4,36 @@ hasClass = (node, search) ->
...
-setupRemoteFromTarget = (target, httpRequestType, urlAttribute, form = null) ->
-  httpUrl = target.getAttribute(urlAttribute)
+setupRemoteFromTarget = (target, httpRequestType, form = null) ->
+  httpUrl = target.getAttribute('href') || target.getAttribute('action')
 
-  throw new Error("Turbograft developer error: You did not provide a URL ('#{urlAttribute}' attribute) for tg-remote") unless httpUrl
+  throw new Error("Turbograft developer error: You did not provide a URL ('data-tg-#{urlAttribute}' attribute) for data-tg-remote") unless httpUrl
 
-  if target.getAttribute("remote-once")
+  if target.getAttribute("data-tg-remote-once") || target.getAttribute("remote-once")
```
DrewMartin: Not a fan of all the `getAttribute('attr') || getAttribute('data-tg-attr')`

How about a `getTGAttribute` method that does
```coffeescript
getTGAttribute(node, attr): ->
  node.getAttribute("data-tg-#{attr}") || node.getAttribute(attr)
```

and then same for remove

DrewMartin: Though because of attributes like `tg-remote` this wouldn't quite work.

Maybe something like
```coffeescript
getTGAttribute(node, attr): ->
  tgAttr = if attr.test[0...3] == 'tg-'
    "data-#{attr}"
  else
    "data-tg-#{attr}"

  node.getAttribute(tgAttr) || node.getAttribute(attr)
```

DrewMartin: Benefits: If we do deprecate the non-data attributes then there's only a couple places that really need to be cleaned up

nwtn: I'm cool with switching to a function like and marking the old attribute names as deprecated in the readme. I'll push that up.

nwtn:
bumped the version to 0.2, updated docs to deprecate old attr names, added some methods to clean up all the `||` bits. @DrewMartin @qq99 whaddya think?

lib/assets/javascripts/turbograft/initializers.coffee
```diff
@@ -1,37 +1,56 @@
...
+    "data-tg-#{attr}"
+
+getTGAttribute = (node, attr) ->
+  tgAttr = tgAttribute(attr)
+  node.getAttribute(tgAttr) || node.getAttribute(attr)
+
+removeTGAttribute = (node, attr) ->
+  tgAttr = tgAttribute(attr)
+  node.removeAttribute(tgAttr)
+  node.removeAttribute(tgAttr)
```
DrewMartin: One of these lines should be `node.removeAttribute(attr)`

lib/assets/javascripts/turbograft/initializers.coffee
```diff
@@ -1,37 +1,56 @@
...
+
+tgAttributeExists = (node, attr) ->
```
DrewMartin: I don't think this method is needed. We can just call `getTGAttribute(node, attr)?`

lib/assets/javascripts/turbograft/initializers.coffee
nwtn: ack good catch!

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -215,7 +215,7 @@ class window.Turbolinks
         else
           refreshedNodes.push(newNode)
 
-      else if existingNode.getAttribute("refresh-always") == null
+      else if existingNode.getAttribute("data-tg-refresh-always") == null && existingNode.getAttribute("refresh-always") == null
```
qq99: `getTGAttribute`?

lib/assets/javascripts/turbograft/initializers.coffee
nwtn: i tried that initially, but it was giving me errors in some of the tests

DrewMartin: And if not, it should probably call `hasAttribute`

lib/assets/javascripts/turbograft/turbolinks.coffee
DrewMartin: Can we make the attribute methods global and use them here as well?

Bonus to use it in query selectors as well to generate the attribute selector

nwtn:
Thanks for the feedback! Will address the remaining bits tomorrow

nwtn:
@DrewMartin @qq99 feedback addressed

qq99:
:+1:

nwtn:
gave it a test run in core at @qq99’s suggestion and didn't run into any issues!

lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -103,7 +103,7 @@ class TurboGraft.Remote
   _enabledInputs: (form) ->
     selector = "input:not([type='reset']):not([type='button']):not([type='submit']):not([type='image']), select, textarea"
     inputs = Array::slice.call(form.querySelectorAll(selector))
-    disabledNodes = Array::slice.call(form.querySelectorAll("[tg-remote-noserialize]"))
+    disabledNodes = Array::slice.call(form.querySelectorAll("[data-tg-remote-noserialize], [tg-remote-noserialize]"))
```
DrewMartin: this could use querySelectorAllTGAttribute

lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -128,7 +128,7 @@ class TurboGraft.Remote
       Page.visit(redirect, reload: true)
       return
 
-    unless @initiator.hasAttribute('tg-remote-norefresh')
+    unless @initiator.hasAttribute('data-tg-remote-norefresh') || @initiator.hasAttribute('tg-remote-norefresh')
```
DrewMartin: And this could use hasTGAttribute

nwtn:
dang thought i got them all. should be fixed now, thanks.

DrewMartin:
:ship:

PR #109, opened 2015-12-06
Make redirect_via_turbolinks_to protected

EiNSTeiN-:
I am trying to ensure all public methods are routable, and this one is public but not routable in Shopify.

@pushrax @DrewMartin

### DIFF
--- a/lib/turbograft/redirection.rb
+++ b/lib/turbograft/redirection.rb
@@ -4,6 +4,7 @@ module TurboGraft
   module Redirection
     extend ActiveSupport::Concern
 
+    private
     def redirect_via_turbolinks_to(url = {}, response_status = {})
       redirect_to(url, response_status)
 


DrewMartin:
Never touched the ruby side of turbograft
@qq99 @Thibaut

lib/turbograft/redirection.rb
```diff
@@ -4,6 +4,7 @@ module TurboGraft
   module Redirection
     extend ActiveSupport::Concern
 
+    protected
```
Thibaut: I'd use `private`

Thibaut:
:+1:

lib/turbograft/redirection.rb
EiNSTeiN-: done

EiNSTeiN-:
I'm not sure what's up with the failing tests?

PR #114, opened 2016-05-19
Fix TG from removing nodes using data prefix on refresh-always

justinthec:
Proposed fix to: https://github.com/Shopify/shopify/issues/67947

Reasoning for Multiline conditional:
 * Nowhere to store the `getTGAttribute` calls ahead of time since this is in an `if... else if...` block. If I evaluate them before the `if... else if...` block then they could be evaluated but not used if the `if` case was entered.
 * Long lines are bad.
 * It's got some documentation around it: https://github.com/jashkenas/coffeescript/issues/966

@qq99 @nwtn

### DIFF
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -215,7 +215,8 @@ class window.Turbolinks
         else
           refreshedNodes.push(newNode)
 
-      else if TurboGraft.getTGAttribute(existingNode, "refresh-always") == null
+      else if (TurboGraft.getTGAttribute(existingNode, "refresh-always") == null &&
+               TurboGraft.getTGAttribute(existingNode, "data-tg-refresh-always") == null)
         removeNode(existingNode)
 
     refreshedNodes


qq99:
I don't think this is the right fix? This will still be doing the `"" || null` I'm pretty sure

I think I would change https://github.com/Shopify/turbograft/blob/master/lib/assets/javascripts/turbograft.coffee#L21-L23 to use `node.hasAttribute` a la https://github.com/Shopify/shopify/issues/67947#issuecomment-203614582

The line changed in this PR would then look like

```coffee
else if !TurboGraft.getTGAttribute(existingNode, "refresh-always") 
```

dfmcphee:
So interestingly enough, I just added `data-tg-refresh="next-nav-avatar"` to the node and then the problem goes away. I'm not sure if there is still another problem with turbograft, but adding a refresh key seems to resolve the issue we saw.

qq99:
@dfmcphee can you verify if putting just `refresh-always` (no `data-` version) resolves the issue?

dfmcphee:
@qq99 yep, that also resolves the issue. very strange

qq99:
So it's definitely [this](https://github.com/Shopify/turbograft/blob/master/lib/assets/javascripts/turbograft.coffee#L14):

```
node.getAttribute(tgAttr) || node.getAttribute(attr)
```
=>
```
node.getAttribute('tg-refresh-always') || node.getAttribute('refresh-always')
```
=>
```
"" || null
```
which actually evaluates to `null`

which makes [this conditional](https://github.com/Shopify/turbograft/blob/7a1fd5212046d1fac407ae3a05be8488dc7314bd/lib/assets/javascripts/turbograft/turbolinks.coffee#L218-L219) return true, whereas before data- attributification it would have been false

I think https://github.com/Shopify/turbograft/pull/114#issuecomment-220212217 is the solution, just hoping that chaning `getAttribute` -> `hasAttribute` in `hasTGAttribute` doesn't cause other issues (it really shouldn't)

justinthec:
Let's continue the discussion on @qq99 's PR:

https://github.com/Shopify/turbograft/issues/116

Issue #118, opened 2016-06-21
error handler often attempts to access an undefined object


GoodForOneFare:
I think this is happening in `TurboGraft.Remote#onError`.  Looks like the exceptions are caused by trying to do `@opts.fail?` when `@opts` is not defined.  I'm not sure how that's possible, given that `@opts` is a constructor assignment.

Shopify's bugsnag shows 18,500+ exceptions since May 26th, so I'm guessing it's related to @justinthec's https://github.com/Shopify/turbograft/commit/dafb3a7f2ee851b64a69c22086ddb6c3c9b9e00a changes.

justinthec:
[Link to BugSnag Event Collection](https://bugsnag.com/shopify/shopify-js/errors/57473f3fbe3f29e6a6051bbc?filters%5Bevent.since%5D%5B%5D=30d)

justinthec:
My PR was merged the [25th](https://github.com/Shopify/turbograft/pull/117#event-672188164) and Twine was bumped on core to use it on [May 26th](https://github.com/Shopify/shopify/pull/73153#event-673511459) so it's 100% related to my changes.

I'll take a look at this.

GoodForOneFare:
Today, an `Uncaught TypeError: window.scrollTo is not a function` error appeared.  It's occurring in `Turbolinks#resetScrollPosition`.  

The kicker is that the error is from `window.scrollTo` not being present in Chrome.  That's crazy.  I've no idea if it's related to this issue, but it seems odd that two "things that should exist"... aren't.

justinthec:
The `@opts.fail?` errors seem to have gone away.

GoodForOneFare:
Fixed by dem fat arrows in https://github.com/Shopify/turbograft/pull/119/files.

PR #119, opened 2016-06-22
Bind this.onError for XHR requests in TurboGraft.Remote

davidcornu:
I have a feeling this is causing https://bugsnag.com/shopify/shopify-js/errors/5666f3c78f2356495b6fe53f?filters%5Bevent.since%5D%5B%5D=30d.

r @qq99 
cc @arapehl

### DIFF
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -116,7 +116,7 @@ class TurboGraft.Remote
       enabledInputs.push(input)
     enabledInputs
 
-  onSuccess: (ev) ->
+  onSuccess: (ev) =>
     @opts.success?()
 
     xhr = ev.target
@@ -145,7 +145,7 @@ class TurboGraft.Remote
         Page.refresh
           response: xhr
 
-  onError: (ev) ->
+  onError: (ev) =>
     @opts.fail?()
 
     xhr = ev.target
--- a/test/javascripts/remote_test.coffee
+++ b/test/javascripts/remote_test.coffee
@@ -67,6 +67,53 @@ describe 'Remote', ->
       assert.equal "/foo/bar", request.url
       assert.equal "POST", request.method
 
+  describe 'callbacks', ->
+
+    it 'will call options.fail() on HTTP failures', (done) ->
+      server = sinon.fakeServer.create()
+      server.respondWith("POST", "/foo/bar",
+            [422, { "Content-Type": "text/html" },
+             '<div id="foo" refresh="foo">Error occured</div>'])
+      remote = new TurboGraft.Remote
+        httpRequestType: "POST"
+        httpUrl: "/foo/bar"
+        refreshOnError: "foo"
+        fail: done
+      , @initiating_target
+      remote.submit()
+
+      server.respond()
+
+    it 'will call options.fail() on XHR failures', (done) ->
+      xhr = sinon.useFakeXMLHttpRequest()
+
+      remote = new TurboGraft.Remote
+        httpRequestType: "POST"
+        httpUrl: "/foo/bar"
+        refreshOnError: "foo"
+        fail: done
+      , @initiating_target
+
+      simulateError = ->
+        listener(target: remote.xhr) for listener in remote.xhr.eventListeners.error
+
+      setTimeout(simulateError, 0)
+
+    it 'will call options.success() on success', (done) ->
+      server = sinon.fakeServer.create()
+      server.respondWith("POST", "/foo/bar",
+            [200, { "Content-Type": "text/html" },
+             '<div>Hey there</div>'])
+      remote = new TurboGraft.Remote
+        httpRequestType: "POST"
+        httpUrl: "/foo/bar"
+        refreshOnError: "foo"
+        success: done
+      , @initiating_target
+      remote.submit()
+
+      server.respond()
+
   describe 'TurboGraft events', ->
 
     beforeEach ->
@@ -74,7 +121,6 @@ describe 'Remote', ->
 
     afterEach ->
       @refreshStub.restore()
-
 
     it 'allows turbograft:remote:init to set a header', ->
       $(@initiating_target).one "turbograft:remote:init", (event) ->
--- a/test/javascripts/remote_test.coffee
+++ b/test/javascripts/remote_test.coffee
@@ -8,7 +8,7 @@ describe 'Remote', ->
   describe 'HTTP methods', ->
 
     it 'will send a GET with _method=GET', ->
-      server = sinon.fakeServer.create();
+      server = sinon.fakeServer.create()
       remote = new TurboGraft.Remote
         httpRequestType: "GET"
         httpUrl: "/foo/bar"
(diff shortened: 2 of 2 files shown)


lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -33,7 +33,7 @@ class TurboGraft.Remote
       triggerEventFor 'turbograft:remote:start', @initiator,
         xhr: xhr
 
-    xhr.addEventListener 'error', @onError
+    xhr.addEventListener 'error', @onError.bind(@)
```
qq99: can you do `.bind(this)`?

qq99:
Looks like it might address a lot of https://github.com/Shopify/turbograft/issues/118

cc @Shopify/tnt

davidcornu:
By the looks of things, the `success` and `fail` options of `TurboGraft.Remote` don't have any test coverage. I'll add some.

lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -33,7 +33,7 @@ class TurboGraft.Remote
       triggerEventFor 'turbograft:remote:start', @initiator,
         xhr: xhr
 
-    xhr.addEventListener 'error', @onError
+    xhr.addEventListener 'error', @onError.bind(this)
```
lemonmade: More CoffeeScript-ey solution is to use `=>` in the method definition (which binds the method in the constructor).

lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -116,7 +116,7 @@ class TurboGraft.Remote
       enabledInputs.push(input)
     enabledInputs
 
-  onSuccess: (ev) ->
+  onSuccess: (ev) =>
```
davidcornu: Figured I'd switch this to a fat arrow too seen as it's always used as a callback

test/javascripts/remote_test.coffee
```diff
@@ -67,6 +67,53 @@ describe 'Remote', ->
...
+
+      remote = new TurboGraft.Remote
+        httpRequestType: "POST"
+        httpUrl: "/foo/bar"
+        refreshOnError: "foo"
+        fail: done
+      , @initiating_target
+
+      simulateError = ->
+        listener(target: remote.xhr) for listener in remote.xhr.eventListeners.error
```
davidcornu: This was the only way I could find to simulate a `xhr.onError` event

davidcornu:
@qq99 & @lemonmade - round 2 please 😁

test/javascripts/remote_test.coffee
```diff
@@ -67,6 +67,53 @@ describe 'Remote', ->
...
+    it 'will call options.fail() on HTTP failures', (done) ->
+      server = sinon.fakeServer.create()
+      server.respondWith("POST", "/foo/bar",
+            [422, { "Content-Type": "text/html" },
+             '<div id="foo" refresh="foo">Error occured</div>'])
+      remote = new TurboGraft.Remote
+        httpRequestType: "POST"
+        httpUrl: "/foo/bar"
+        refreshOnError: "foo"
+        fail: done
```
lemonmade: If this never failed, would this test just timeout? Might be worth adding the success case that throws a meaningful error.

test/javascripts/remote_test.coffee
lemonmade: Yeah, I can't really think of how to test this nicely either :S

test/javascripts/remote_test.coffee
```diff
@@ -67,6 +67,53 @@ describe 'Remote', ->
...
+      , @initiating_target
+
+      simulateError = ->
+        listener(target: remote.xhr) for listener in remote.xhr.eventListeners.error
+
+      setTimeout(simulateError, 0)
```
lemonmade: Why do you have to delay this?

test/javascripts/remote_test.coffee
```diff
@@ -67,6 +67,53 @@ describe 'Remote', ->
...
+
+    it 'will call options.success() on success', (done) ->
+      server = sinon.fakeServer.create()
+      server.respondWith("POST", "/foo/bar",
```
lemonmade: whitespace is a little wonky on these, I would probably do one arg per line.

test/javascripts/remote_test.coffee
```diff
@@ -67,6 +67,53 @@ describe 'Remote', ->
...
+            [200, { "Content-Type": "text/html" },
+             '<div>Hey there</div>'])
+      remote = new TurboGraft.Remote
+        httpRequestType: "POST"
+        httpUrl: "/foo/bar"
+        refreshOnError: "foo"
+        success: done
```
lemonmade: SO I guess, if these tests don't time out that's the success case. In addition to the "adding an error for the fail case" suggestion I made above, maybe use a spy for the success case callback, and check that the spy was called (again, just to be more explicit)?

test/javascripts/remote_test.coffee
qq99: Interesting

qq99:
Just going back to look at https://github.com/Shopify/turbograft/commit/dafb3a7f2ee851b64a69c22086ddb6c3c9b9e00a#diff-58f1b55d4025fab95951247202574f55L156 to analyze some more...

Looks like previously, `@refreshOnError` would have been undefined in some cases, and so it'd just do a general `triggerEventFor` and that'd be the end of it. It also looks like the network-level failures were/are the ones with problem (lacking proper `this`), combined with trying to do multi-level object lookup now

@davidcornu LGTM let me know when you'd like me to 🎩 this against core

test/javascripts/remote_test.coffee
davidcornu: To guarantee async. XHR error callbacks would never be called within the same tick.

davidcornu:
> It also looks like the network-level failures were/are the ones with problem (lacking proper `this`), combined with trying to do multi-level object lookup now

Yeah that was my conclusion.

test/javascripts/remote_test.coffee
davidcornu: @lemonmade was adhering to the style in  the rest of the file. Should I update all of it or just this one?

test/javascripts/remote_test.coffee
davidcornu: The error scenario is a timeout for both of these cases. I can't think of a way to give a better error message though as there's no way to hook into the timeout behaviour. What I want is

```javascript
it('calls the callback within 2 seconds', function(done){
  var failureTimeout = setTimeout(function(){
    done(new Error("callback was not called"));
  }, 2000);

  doSomethingAsync(function(){
    clearTimeout(failureTimeout);
    done();
  });
});
```

but that seems like a lot of code to make the error message better.

test/javascripts/remote_test.coffee
lemonmade: If it's the same everywhere, KKSO 👍

test/javascripts/remote_test.coffee
lemonmade: talked on Slack, I'm good with the current version 👍

davidcornu: Talked to @lemonmade on Slack. Timeout is acceptable in this case.

davidcornu:
@qq99 should be good to go. LMK if you want any help with 🎩 .

qq99:
Tested it in a few scenarios (order actions, some 422s) and looks good 🎩

qq99:
🚢

PR #121, opened 2016-08-03
Fix deprecations and tests

gmalette:
This PR fixes deprecation warnings for Rails 5

### DIFF
--- a/lib/turbograft.rb
+++ b/lib/turbograft.rb
@@ -21,8 +21,8 @@ module TurboGraft
         Config.controllers.each do |klass|
           klass.constantize.class_eval do
             include XHRHeaders, Cookies, XDomainBlocker, Redirection
-            before_filter :set_xhr_redirected_to, :set_request_method_cookie
-            after_filter :abort_xdomain_redirect
+            before_action :set_xhr_redirected_to, :set_request_method_cookie
+            after_action :abort_xdomain_redirect
           end
         end
 
@@ -36,7 +36,7 @@ module TurboGraft
 
       ActiveSupport.on_load(:action_view) do
         (ActionView::RoutingUrlFor rescue ActionView::Helpers::UrlHelper).module_eval do
-          include XHRUrlFor
+          prepend XHRUrlFor
         end
       end unless RUBY_VERSION =~ /^1\.8/
     end
--- a/test/example/app/controllers/application_controller.rb
+++ b/test/example/app/controllers/application_controller.rb
@@ -1,8 +1,6 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery
 
-  skip_before_filter :authenticate, :only => :unauthenticated
-
   def test
     head :ok
   end
--- a/lib/turbograft/xhr_url_for.rb
+++ b/lib/turbograft/xhr_url_for.rb
@@ -3,13 +3,9 @@ module TurboGraft
   # option by using the X-XHR-Referer request header instead of the standard Referer
   # request header.
   module XHRUrlFor
-    def self.included(base)
-      base.alias_method_chain :url_for, :xhr_referer
-    end
-
-    def url_for_with_xhr_referer(options = {})
+    def url_for(options = {})
       options = (controller.request.headers["X-XHR-Referer"] || options) if options == :back
-      url_for_without_xhr_referer options
+      super(options)
     end
   end
 end


gmalette:
@nwtn @qq99

test/example/app/controllers/application_controller.rb
```diff
@@ -1,8 +1,6 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery
 
-  skip_before_filter :authenticate, :only => :unauthenticated
```
gmalette: I have no idea where this came from, but it was breaking because `:authenticate` doesn't exist

lib/turbograft.rb
```diff
@@ -36,7 +36,7 @@ def referer
 
       ActiveSupport.on_load(:action_view) do
         (ActionView::RoutingUrlFor rescue ActionView::Helpers::UrlHelper).module_eval do
-          include XHRUrlFor
+          prepend XHRUrlFor
         end
       end unless RUBY_VERSION =~ /^1\.8/
```
gmalette: Given we only test on Ruby 2.2, I was planning on bumping the gemspec to that and release `0.3.0`

qq99:
cc @Shopify/tnt

gmalette:
fyi @Shopify/scripts

nwtn:
i don’t think i'll be very useful in reviewing this, so leaving it in the hands of the fine tnt folks

lib/turbograft.rb
```diff
@@ -36,7 +36,7 @@ def referer
 
       ActiveSupport.on_load(:action_view) do
         (ActionView::RoutingUrlFor rescue ActionView::Helpers::UrlHelper).module_eval do
-          include XHRUrlFor
+          prepend XHRUrlFor
```
kurtfunai: Just general curiosity here - why the change from include to prepend?

gmalette: **TL, DR: ** `prepend` was added in Ruby 2, removing the need for `alias_method_chain`, which will be removed in Rails 5.

--- 
Say you want collections in Ruby  `module Identifiable` that gets included in other classes to extend the functionality the `id` method to also not return the name of the class.

```ruby
module Identifiable
  def id
    self.class_name + "." + super
  end
end

class User
  include Identifiable

  def id 
    primary_key
  end
end
```

## `include`

If you did this this way, Ruby will make `Identifiable` an ancestor of `User`. The ancestor chain of `User` would be 
```
[User, Identifiable, ...]
```

Notice `User` is first in the chain, thus calling `user.id` would simply return the primary key. Because of this, Rails implemented `alias_method_chain`, which you can use similarly to `alias_method_chain :id, :class_name`. 

This will do two things:
1 ) Rename the method `id` in `User` to `id_without_class_name`
2) Define a new `id` method like such:

```ruby
def id
  id_with_class_name
end
```

You would then define the `id_with_class_name` method in the inherited module (`Identifiable`) to do what you want, and instead of calling `super`, you would call `id_without_class_name`. This was basically a way to inject a more precise inheritance in the chain of the object when the language didn't support it. The dispatch diagram looks like something like this

```
+--------+          +----------------+
|        |          |                |
|  User  |          |  Identifiable  |
|        |          |                |
+--------+          +----------------+

  id  +------------->  id_with_class_name
                             +
                             |
  id_without_class_name <----+
  (originally called id)
```

## `prepend`

Using `prepend` flips things over. It prepends the module in the chain, so you get 

```ruby
[Identifiable, User, ...]
```

When you call `id`, it will hit the definition in `Identifiable` first, which calls `super`, which then finds the definition in `User`, resulting in the intended behaviour. 

The dispatch is much simpler to grok and debug

```
+---------------+   +--------+
|               |   |        |
| Identifiable  |   |  User  |
|               |   |        |
+---------------+   +--------+

  id   +----------->  id
```

kurtfunai: Thank you for such a thorough and well written explanation 😃  🎆 

Makes a lot of sense.

qq99: Really nice explanation, thanks!

qq99:
Anyone available to 🎩 on admin?

gmalette:
@Shopify/tnt halp

test/example/app/controllers/application_controller.rb
gmalette: So this is because Rails 5 doesn't let you skip actions that are not applied.

gmalette: Must've been legacy and someone forgot to remove it some time ago

justinthec:
I'll give it a :tophat: today

justinthec:
🎩  💚 
Tested on OSX: Chrome, Firefox, and Safari

gmalette:
🚢 ?

qq99:
:shipit: imo

gmalette:
fyi @rafaelfranca

lib/turbograft.rb
```diff
@@ -36,7 +36,7 @@ def referer
 
       ActiveSupport.on_load(:action_view) do
         (ActionView::RoutingUrlFor rescue ActionView::Helpers::UrlHelper).module_eval do
```
phoet: do you actually need to `module_eval` anything here? AFAIR `Module#include` and `Module#prepend` were made public in v2.1.

justinthec:
image

```
Teaspoon running default suite at http://127.0.0.1:53984/teaspoon/default
rake aborted!
Errno::EINVAL: Invalid argument - bind(2) for "fe80::4001:aff:fef0:30" port 7055
```

@gmalette

lib/turbograft.rb
rafaelfranca: yeah, I think it is not needed.

rafaelfranca:
Thanks!

gmalette:
@justinthec is that related to my changes?

justinthec:
Most likely, it only became red after your commit.

gmalette:
Any idea what would cause that?

PR #122, opened 2016-08-10
Add minimum Ruby version

gmalette:
This PR bumps the Ruby dependency of TG to Ruby 2.0. This is required for `prepend`.

It also addresses [@phoet's comment](https://github.com/Shopify/turbograft/pull/121#discussion_r74129271).

PR #125, opened 2016-08-11
prepend is public since 2.1

phoet:
since the method has been made public, we can invoke it without any `moduel_eval`ing

refs https://github.com/Shopify/turbograft/pull/121#discussion_r74129271


gmalette:
See #122

PR #126, opened 2016-08-11
Redo patch to fix deprecations and tests

gmalette:


.travis.yml
```diff
@@ -2,10 +2,8 @@ language: ruby
 rvm:
   - '2.2.2'
 before_script:
+  - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
+  - sleep 3
```
gmalette: @justinthec this PR is identical to the one I merged this week, and is also green. 

The above are recommendations from Travis wrt xvfb 
Below is because it obviously doesn't work as expected so we were just spinning up 3 builds for no reason

justinthec:
Thanks for fixing this @gmalette !

PR #127, opened 2016-08-11
add dev.yml to make life like, so much easier

TheMallen:
After much trials and tribulations trying to get this project to run on my and a few others machines, I realized that was silly and it should just have a `dev.yml`.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -1,0 +1,9 @@
+---
+name: turbograft
+up:
+- ruby: 2.3.1
+- bundler
+commands:
+  test: bundle exec rake test
+  js-test: bundle exec teaspoon
+  server: ./server


TheMallen:
@justinthec @darrenhebner @Shopify/tnt

I merge this yes? Plis.

dev.yml
```diff
@@ -0,0 +1,9 @@
+---
+name: turbograft
+up:
+- ruby: 2.3.1
+- bundler
+commands:
+  test: bundle exec rake test
```
edward: In Shopify/shopify this is done by `bin/rake` instead of `bundle exec rake`: https://github.com/Shopify/shopify/blob/master/dev.yml#L53

Does that do the same thing?

dev.yml
```diff
@@ -0,0 +1,9 @@
...
+  js-test: bundle exec teaspoon
+  server: ./server
```
edward: Is specifying a path necessary?

edward:
Looks good other than my comments.

edward:
Did you take a look at https://vault.shopify.com/dev/yml#dev-yml_up and `dev help init` / `dev init`?

TheMallen:
Yes I generated this using `dev init` and then added the `server` command. You need the path because it's referencing a local script to execute.

dev.yml
TheMallen: `bundle exec rake` is better I because it will always run in the context of your current project's bundle. If you have system gems `rake` could use those instead.

justinthec:
👍

PR #128, opened 2016-08-11
add js test info to readme and add environment var for teaspoon driver

TheMallen:
This PR adds more information about running tests to the readme.

This PR also adds a link to some troubleshooting information about why selenium will probably explode when you try to do this, and how to not have it explode.


TheMallen:
I wasn't able to get `selenium` working on my machine, so I added an environment variable to allow you to use whatever teaspoon driver you want and switched the default to phantomjs.

edward:
Looks great

justinthec:
My tests run locally now after setting the env var to phantomjs! 🎉

PR #123, opened 2016-08-10
Update <head> with new assets on page navigation instead of full page refresh

justinthec:
>Put a PR up anyway
image

@darrenhebner @TheMallen @Shopify/tnt 

@GoodForOneFare @edward 

~~**Note:** JS tests are order dependant due to the DOM being shared and not cleaned up between tests however clearing the body causes more assertion errors. We will continue to look into properly fixing this order dependency issue but for now, we've moved our new tests to a location where it does not cause conflicts.~~

PR #131, opened 2016-08-25
Replaced all assert while loops from selenium tests with proper Capybara assertions

justinthec:
There were a couple cases of
```ruby
has_text = false
while !has_text
  has_text = page.assert_text("something")
  sleep 1
assert_equal something, 'something'
```
in our test suite and they were causing tests to flake in CI since the `while` loop can sometimes pass the assertion on the first iteration, or fail numerous times in subsequent assertions before exiting the loop.

We fixed them to use `has_text` instead of `assert_text` in https://github.com/Shopify/turbograft/pull/123 and this prevented flakiness since the test would not fail when `has_text` returned false.

However @GoodForOneFare pointed out that Capybara actually has the [built in behaviour](https://github.com/jnicklas/capybara/blob/master/README.md#asynchronous-javascript-ajax-and-friends) to keep repeating a failing assertion for 2 seconds before it gives up and actually produces a failure. The 2 second grace period can be adjusted in the config or on a per test basis by passing in a `wait` argument to the assertion.

This PR leverages these Capybara assertions to remove the `while` loops completely from all of our tests and the symptom of flakiness.

For review:
@Shopify/tnt @TheMallen

### DIFF
--- a/test/browser/full_page_refresh_test.rb
+++ b/test/browser/full_page_refresh_test.rb
@@ -40,28 +40,18 @@ class FullPageRefreshTest < ActionDispatch::IntegrationTest
 
   test "going to a URL that will error 500, and hitting the browser back button, we see the correct page (and not the 500)" do
     click_link "I will throw an error 500"
-    has_text = false
-    while !has_text
-      has_text = page.assert_text('Error 500!')
-      sleep 1
-    end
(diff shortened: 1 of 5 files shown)


TheMallen:
image

PR #132, opened 2016-08-26
Fix safari no styles

TheMallen:
This PR fixes the `sometimes the app breaks` case where a user navigates several times within a TG application using `0.3.0`.

The breakages were the result of `Turbohead#update` missing the upstream document's css when processing it. This only occurred randomly (somewhere between 1/10 and 1/20 times) and only on safari. It seems to be the result of:

```
extractTrackedAssets = (doc) ->
  for node in doc.head.children when node.dataset.turbolinksTrack?
    node
```

Where the result of extractTrackedAssets would be different than the new implementation:

```
extractTrackedAssets = (doc) ->
  return [].slice.call(doc.querySelectorAll('[data-turbolinks-track]'))
```

However, only sometimes, and only on safari. Debugging through or adding `console.log` to the loop steps would make the function perform as expected.

Since `querySelector` does not give you a live collection, it seems to be some strange browser level behaviour involving live collections of DOM elements. Similar flakiness was also addressed previously in  https://github.com/Shopify/turbograft/pull/93.

~~In addition, there was some concern that we were not testing large numbers of subsequent navigations (though this is not specifically the issue as far as I can tell), so I've included additional tests for this case.~~

### DIFF
--- a/lib/assets/javascripts/turbograft/turbohead.coffee
+++ b/lib/assets/javascripts/turbograft/turbohead.coffee
@@ -26,8 +26,7 @@ updateScriptTags = (activeDocument, newScripts, callback) ->
   )
 
 extractTrackedAssets = (doc) ->
-  for node in doc.head.children when node.dataset.turbolinksTrack?
-    node
+  return [].slice.call(doc.querySelectorAll('[data-turbolinks-track]'))
 
 filterForNodeType = (nodeType) ->
   (node) -> node.nodeName == nodeType
--- a/lib/assets/javascripts/turbograft/turbohead.coffee
+++ b/lib/assets/javascripts/turbograft/turbohead.coffee
@@ -26,7 +26,7 @@ updateScriptTags = (activeDocument, newScripts, callback) ->
   )
 
 extractTrackedAssets = (doc) ->
-  return [].slice.call(doc.querySelectorAll('[data-turbolinks-track]'))
+  [].slice.call(doc.querySelectorAll('[data-turbolinks-track]'))
 
 filterForNodeType = (nodeType) ->
   (node) -> node.nodeName == nodeType
--- a/test/javascripts/turbolinks_test.coffee
+++ b/test/javascripts/turbolinks_test.coffee
@@ -64,7 +64,7 @@ describe 'Turbolinks', ->
 
   visit = ({options, url}, callback) ->
     $(document).one('page:load', (event) ->
-      setTimeout((-> callback(event)), 0)
+      setTimeout((-> callback(event) if callback), 0)
     )
     Turbolinks.visit('/' + url, options)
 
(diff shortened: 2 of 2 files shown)


test/javascripts/turbolinks_test.coffee
```diff
@@ -185,6 +185,33 @@ describe 'Turbolinks', ->
           assertLinks(['foo.css'])
           done()
 
+    it 'many subsequent navigations do not break head asset tracking', (done) ->
```
GoodForOneFare: Unless we have a way to run Safari in CI, these don't seem valuable.  They probably give a false sense of confidence to the next people who come along and try to update the `head` query.

lib/assets/javascripts/turbograft/turbohead.coffee
```diff
@@ -26,8 +26,7 @@ updateScriptTags = (activeDocument, newScripts, callback) ->
   )
 
 extractTrackedAssets = (doc) ->
-  for node in doc.head.children when node.dataset.turbolinksTrack?
-    node
+  [].slice.call(doc.querySelectorAll('[data-turbolinks-track]'))
```
GoodForOneFare: Does this do anything that [the original turbolinks fix](https://github.com/turbolinks/turbolinks-classic/blob/master/lib/assets/javascripts/turbolinks.coffee#L352) doesn't do?

TheMallen: It outputs considerable more legible javascript than a coffeescript array comprehension.

test/javascripts/turbolinks_test.coffee
TheMallen: That's fair. I'll wipe them. :)

GoodForOneFare:
Changes LGTM.  Should I 🎩?

TheMallen:
Yep, absolutely 🎩  if you can.

If you point your repo at `0.3.0` on `master` and navigate around for a bit on safari you should lose all your styles. If you do the same with this branch you should be 🆗

lib/assets/javascripts/turbograft/turbohead.coffee
qq99: How about `Array.prototype.slice` instead of `[].slice`? http://stackoverflow.com/questions/9006553/slice-or-array-prototype-slice suggests prototype is more performant, unfortunately jsperf is dead :(

qq99:
Changes LGTM 👍

lib/assets/javascripts/turbograft/turbohead.coffee
WillsonSmith: The performance diffs here are so negligible it doesn't really matter what you choose unless you're going to run this hundreds of thousands of times.

Average of running 1 million instances. Each one of these were run 5 separate times.

| [].slice.call | Array.prototype.slice.call |
| ------- |------------------
| 394.88ms |378.16ms |

That's about 16ms at 1 million times, about 1 frame.

TheMallen: Yeah, at that point it's mostly a matter of whether you think `Array.prototype.slice.call` or `[].slice.call` looks nicer.

qq99: thanks! 👍 as is!

TheMallen:
Cool I think this is g2g then

PR #133, opened 2016-08-29
Fix partial replace head tracking

TheMallen:
This PR fixes partial replaces inadvertently triggering head asset updates. This was usually not causing issues, but in some cases could break stuff.

I should note that this was partially to blame for the spike in `url.absolute` errors. Submitting a remote form would call: 

```
        Page.refresh
          response: xhr
          onlyKeys: @refreshOnSuccess
```

`Page.refresh` would then trigger

```
    options.partialReplace = true
    options.onLoadFunction = callback

    xhr = options.response
    delete options.response
    Turbolinks.loadPage null, xhr, options
```

Which would then try to update the head assets, with an error callback that looked like:

```
onHeadUpdateError = ->
          Turbolinks.fullPageNavigate(url.absolute)
```

Where `url` is null. 

Preventing head asset updates on partial request should fix the new instances of this error that all occurred with a `tg-remote` related last event.

The old instances of this error are a bit more insidious and will be addressed in an upcoming PR.

### DIFF
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -132,15 +132,16 @@ class window.Turbolinks
     options.updatePushState ?= true
     if upstreamDocument = processResponse(xhr)
       reflectNewUrl url if options.updatePushState
-      unless options.partialReplace
-        return new TurboHead(document, upstreamDocument).update(
+      if options.partialReplace
+        updateBody(upstreamDocument, xhr, options)
+      else
+        new TurboHead(document, upstreamDocument).update(
           onHeadUpdateSuccess = ->
             updateBody(upstreamDocument, xhr, options)
           ,
           onHeadUpdateError = ->
             Turbolinks.fullPageNavigate(url.absolute)
         )
-      return updateBody(upstreamDocument, xhr, options)
     else
       triggerEvent 'page:error', xhr
       Turbolinks.fullPageNavigate(url.absolute) if url?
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -130,27 +130,31 @@ class window.Turbolinks
...
+      unless options.partialReplace
+        return new TurboHead(document, upstreamDocument).update(
+          onHeadUpdateSuccess = ->
+            updateBody(upstreamDocument, xhr, options)
+          ,
+          onHeadUpdateError = ->
+            Turbolinks.fullPageNavigate(url.absolute)
+        )
+      return updateBody(upstreamDocument, xhr, options)
+    Turbolinks.fullPageNavigate(url.absolute)
```
qq99: your logic changes LGTM
My only comment would be that I prefer the style of explicit if/else rather than `return` to terminate control flow

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -130,27 +130,31 @@ class window.Turbolinks
   @loadPage: (url, xhr, options = {}) ->
     triggerEvent 'page:receive'
     options.updatePushState ?= true
-    if upstreamDocument = processResponse(xhr, options.partialReplace)
+    if upstreamDocument = processResponse(xhr)
```
GoodForOneFare: Think we can 🔥 the `partial` parameter from `processReponse` now.

GoodForOneFare:
Makes sense 👍

lib/assets/javascripts/turbograft/turbolinks.coffee
TheMallen: yass

TheMallen:
I'll 🚢  this when it's done on CI

TheMallen:
Had to rebase on top of the other two prs

PR #135, opened 2016-08-29
bump version

TheMallen:


GoodForOneFare:
Tophatting in Safari, the only problem I've seen so far is momentary unstyling of the page when a CSS asset fingerprint changes.
 
css-asset-bump-and-jank

Aside form that, I've no had problems while going through: products, orders, transfers, customers, reports, dashboard, online store, customize theme (refreshes, as expected), edit theme (refreshes, as expected), inline form updates, full page form updates, asset fingerprint updates, blog posts, store pages, store navigation, apps, general settings, shipping (+ sub-pages), gift cards, notifications, customize email.

GoodForOneFare:
Tophatting in Firefox, things are the same as Safari.  If I add a CSS style, there's a janky second as Turbohead removes/adds stuff.

Also discovered that clearing cookies then navigating leads to this being displayed:
shopify please log in 2016-08-30 12-38-17

GoodForOneFare:
Tophatting in IE hasn't gone well so far.  I'm downloading a new VM in the hope that it'll make everything better.

qq99:
re: FOUC, do we remove current `some-style-file-fingerprint1.css` before adding `some-style-file-fingerprint2.css` to `<head>`?  I wonder, can we add replacement to head before removing original? Just a musing since I haven't checked the implementation

qq99:
Another good flow to check might be 403s:

- staff member doesn't have access to section of admin
- staff member gets creds revoked mid-use

TheMallen:
I can look into our link graph monster and see if I can mitigate the unstyled page issues.

The cookies thing is interesting.

Let me know how IE goes after vm rebuild.

GoodForOneFare:
Going to _Settings > Checkout_, then clicking _Customize Checkout_ results in this:

shop 1 customize pop shopify 2016-08-30 13-55-25

Verified in Edge + Chrome.  Most likely everything else.  Refreshing the page gets the editor to load, but then the back button stops working (URL changes but checkout editor stays onscreen).

GoodForOneFare:
Tophat in IE11 went surprisingly well.  Tabs in the products index get funky if you change CSS assets and switch tab, but I'm hoping that'll be fixed by the upcoming algo tweaks.

TheMallen:
I'm actually going with the dirt simple implementation I think. I think that going with the simplest possible solution that suits our needs is expedient at this point. I'll see how it pans out shortly. (it's also drastically less code)

PR #137, opened 2016-08-31
Revert to the simpler solution, :(

TheMallen:
image
image

I'm very sad tonight.

The graph was causing flickering on new asset bundles no matter how I refactored it due to stylesheet parsing. Load events on link tags weren't working in test environments or some of the browsers I tried. The only solutions seemed to be extremely verbose and hacky (such as setting a polling function to see if asset definitions were loaded by reaching into style tags, or using other nonsense like that)

As such, I just pulled a big old 🔥  and went with the simplest possible solution that gets us what we need.

On the brightside, I kept things at a nice balance of modular and simple, so it should be relatively easy to enhance our solution if/when we think up better ones. If we can work around test environments and are ok with older browsers (like literally firefox from a couple months ago) exploding, we could have a more robust solution again in a very small PR.

TLDR:
- We only ever insert new head assets
- `data-turbolinks-track-as` and `data-turbolinks-track` are now merged, using `data-turbolinks-track='foo'` is the same as using both of the old attributes together with a `track-as` of foo.
- We don't even care about load events from links. Don't worry bout it.
- Our solution suits our needs better than stealing mainstream Turbolink's because we don't want a full page refresh when we navigate to a page with less `link` tags unless we specify.

Anyway, I'm going to bed now, talk to y'all tommorrow :p

### DIFF
--- a/lib/assets/javascripts/turbograft/turbohead.coffee
+++ b/lib/assets/javascripts/turbograft/turbohead.coffee
@@ -176,7 +176,7 @@ reorderActiveLinks = (activeLinks, upstreamLinks) ->
               document.head.insertBefore(linkClone, currentLink.nextSibling)
               removeLink(linkToMove, startIndex)
               activeLinksCopy.splice(i, 0, linkToMove)
-              triggerEvent('page:after-link-inserted', linkToMove)
+              triggerEvent('page:after-link-inserted', linkClone)
               return
           else
             addNewReorder(linkToMove, currentLink, pendingReorders)
--- a/lib/assets/javascripts/turbograft/turbohead.coffee
+++ b/lib/assets/javascripts/turbograft/turbohead.coffee
@@ -60,18 +60,6 @@ asyncSeries = (tasks, callback) ->
   task = tasks.shift()
   task(-> asyncSeries(tasks, callback))
 
-
-asyncParallel = (tasks, callback) ->
-  tasksRemaining = tasks.length
-  return callback() if tasksRemaining == 0
-
-  done = () ->
-    tasksRemaining--
-    if (tasksRemaining == 0)
-      callback()
-
-  tasks.forEach((task) -> task(done))
-
 insertScriptTask = (activeDocument, scriptNode) ->
   # We need to clone script tags in order to ensure that the browser executes them.
   newNode = activeDocument.createElement('SCRIPT')
--- a/lib/assets/javascripts/turbograft/turbohead.coffee
+++ b/lib/assets/javascripts/turbograft/turbohead.coffee
@@ -1,41 +1,74 @@
+TRACKED_ASSET_SELECTOR = '[data-turbolinks-track]'
+TRACKED_ATTRIBUTE_NAME = 'turbolinksTrack'
+ANONYMOUS_TRACK_VALUE = 'true'
+
 class window.TurboHead
   constructor: (@activeDocument, @upstreamDocument) ->
     @activeAssets = extractTrackedAssets(@activeDocument)
     @upstreamAssets = extractTrackedAssets(@upstreamDocument)
     @newScripts = @upstreamAssets
-      .filter(filterForNodeType('SCRIPT'))
-      .filter(noMatchFor({attribute: 'src', inCollection: @activeAssets}))
+      .filter(attributeMatches('nodeName', 'SCRIPT'))
+      .filter(noAttributeMatchesIn('src', @activeAssets))
 
     @newLinks = @upstreamAssets
-      .filter(filterForNodeType('LINK'))
-      .filter(noMatchFor({attribute: 'href', inCollection: @activeAssets}))
+      .filter(attributeMatches('nodeName', 'LINK'))
+      .filter(noAttributeMatchesIn('href', @activeAssets))
 
-  hasAssetConflicts: () ->
+  hasChangedAnonymousAssets: () ->
+    anonymousUpstreamAssets = @upstreamAssets
+      .filter(datasetMatches(TRACKED_ATTRIBUTE_NAME, ANONYMOUS_TRACK_VALUE))
+    anonymousActiveAssets = @activeAssets
+      .filter(datasetMatches(TRACKED_ATTRIBUTE_NAME, ANONYMOUS_TRACK_VALUE))
+
+    if anonymousActiveAssets.length != anonymousUpstreamAssets.length
+      return true
+
+    anonymousActiveAssets.some(
+      noAttributeMatchesIn(TRACKED_ATTRIBUTE_NAME, anonymousUpstreamAssets)
+    )
+
+  hasNamedAssetConflicts: () ->
     @newScripts
       .concat(@newLinks)
-      .some(hasAssetConflicts(@activeAssets))
+      .filter(noDatasetMatches(TRACKED_ATTRIBUTE_NAME, ANONYMOUS_TRACK_VALUE))
+      .some(datasetMatchesIn(TRACKED_ATTRIBUTE_NAME, @activeAssets))
+
+  hasAssetConflicts: () ->
+    @hasNamedAssetConflicts() || @hasChangedAnonymousAssets()
 
   insertNewAssets: (callback) ->
     updateLinkTags(@activeDocument, @newLinks)
     updateScriptTags(@activeDocument, @newScripts, callback)
 
 extractTrackedAssets = (doc) ->
-  [].slice.call(doc.querySelectorAll('[data-turbolinks-track]'))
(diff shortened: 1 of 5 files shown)


TheMallen:
Caveat,

If you think you know a better solution let me know. I made this PR when I was very sleepy.

lib/assets/javascripts/turbograft/turbohead.coffee
```diff
@@ -4,211 +4,90 @@ class window.TurboHead
...
+    callback
+  )
 
 asyncSeries = (tasks, callback) ->
   return callback() if tasks.length == 0
   task = tasks.shift()
   task(-> asyncSeries(tasks, callback))
 
+
+asyncParallel = (tasks, callback) ->
```
GoodForOneFare: Should this be used somewhere?

TheMallen: Oops.

This is a leftover from when I was still trying to find a way to load/unload the link tags while waiting for all of their events asynchronously. It can be removed for the time being.

GoodForOneFare:
Tophatting in the bundle split branch, and things are looking good to me:
- Named CSS assets cause a page refresh when changed
- Named JS assets cause a page refresh when changed

We still have to fix the expired session and checkout, but that can happen in other PRs.  I also have to figure out tophatting for 4**/5** errors.

lib/assets/javascripts/turbograft/turbohead.coffee
```diff
@@ -4,39 +4,56 @@ class window.TurboHead
   update: (successCallback, failureCallback) ->
```
GoodForOneFare: This function should either:
- Gather a set of `data-turbolinks-track: 'true'` assets, and perform a full page refresh if the list changes
- Fail fast with a, `'true' is no longer supported` exception

The PR in its current form just seems to dynamically load `data-turbolinks-track: true` assets as their fingerprints change.  That's the exact opposite of the old behaviour.

TheMallen: Hmmm, I like the backwards compatibility one. :ok_hand:

lib/assets/javascripts/turbograft/turbohead.coffee
```diff
@@ -4,39 +4,56 @@ class window.TurboHead
...
+      .filter(noMatchFor({attribute: 'href', inCollection: activeAssets}))
 
-updateLinkTags = (activeAssets, upstreamAssets) ->
-  activeLinks = activeAssets.filter(filterForNodeType('LINK'))
-  upstreamLinks = upstreamAssets.filter(filterForNodeType('LINK'))
-  remainingActiveLinks = removeStaleLinks(activeLinks, upstreamLinks)
-  reorderedActiveLinks = reorderActiveLinks(remainingActiveLinks, upstreamLinks)
-  insertNewLinks(reorderedActiveLinks, upstreamLinks)
+    if newScripts.concat(newLinks).some(hasAssetConflicts(activeAssets))
+      return failureCallback()
```
GoodForOneFare: `TurboHead`'s API should be:
```
constructor(activeDocument, upstreamDocument)
boolean hasUpdatedAsset()
insertNewAssets(callback)
```

Then this oddly named var won't be necessary, and the [bizzaro variable creation to simulate named parameters](https://github.com/Shopify/turbograft/blob/bd66d218af5b438baeb3f73ff1ae07f26d342ace/lib/assets/javascripts/turbograft/turbolinks.coffee#L139) code won't be necessary.

lib/assets/javascripts/turbograft/turbohead.coffee
```diff
@@ -4,39 +4,56 @@ class window.TurboHead
...
 
-updateScriptTags = (activeDocument, newScripts, callback) ->
-  asyncSeries(
-    newScripts.map((scriptNode) -> insertScriptTask(activeDocument, scriptNode)),
-    callback
-  )
+    updateTasks = [
+      (done) => updateLinkTags(@activeDocument, newLinks, done),
```
GoodForOneFare: You mentioned leaving this as it is in Slack, but I think it's worth moving the link tags out of the async logic.  Code should reflect existing intent.

lib/assets/javascripts/turbograft/turbohead.coffee
TheMallen: More to simulate passing an inlined named function (coffeescript doesn't let you), but ok.

lib/assets/javascripts/turbograft/turbohead.coffee
TheMallen: Kay.

edward:
I tophatted in latest Chrome and Safari yesterday and things are looking good.

When I introduced extra style assets in the head for a particular section (e.g. Orders/show) by way of a `content_for(:head)` and marked the stylesheet link tag with a `turbolinks_track: true` attribute, I could see that the asset was being added as soon as I reached that section. These styles were continued to be applied as I navigated around the admin.

Updates to the style asset were reflected when I reached the page again.

GoodForOneFare:
@edward to be clear, the final version of this PR will not have the additive CSS behaviour.

edward:
What’s the additive CSS behaviour?

GoodForOneFare:
@edward the final behaviour will look like this:
* Stylesheet is added to a page
* Turbogears sees a new "legacy" tracked asset
* Turbogears refreshes because that's how legacy assets have always behaved

However, if someone adds `turbolinks_track: 'my-awesome-css'`, the sheet will be added sans refresh.

test/javascripts/turbolinks_test.coffee
```diff
@@ -122,243 +122,175 @@ describe 'Turbolinks', ->
...
+        visit url: 'singleScriptInHeadTrackTrue', ->
+          assert(Turbolinks.fullPageNavigate.callCount, 2, 'Should perform two full page refreshes.')
           done()
 
-    it 'inserts link with a new href into existing head on navigation', (done) ->
-      visit url: 'singleLinkInHead', ->
-        assertLinks(['foo.css'])
-        visit url: 'twoLinksInHead', ->
-          assertLinks(['foo.css', 'bar.css'])
+      it 'does not refresh page when tracked node are the same', (done) ->
```
GoodForOneFare: Missing plularization, and vague.  Suggest `when tracked nodes have matching URIs`

test/javascripts/turbolinks_test.coffee
```diff
@@ -122,243 +122,175 @@ describe 'Turbolinks', ->
...
+        visit url: 'singleScriptInHeadTrackTrue', ->
+          assert(Turbolinks.fullPageNavigate.called, 'Should perform a full page refresh.')
           done()
 
-    it 'inserts link with a new href into empty head on navigation', (done) ->
-      visit url: 'noScriptsOrLinkInHead', ->
-        assertLinks([])
-        visit url: 'singleLinkInHead', ->
-          assertLinks(['foo.css'])
+      it 'refreshes page when an extant tracked node is missing', (done) ->
```
GoodForOneFare: Is this equivalent to `when tracked node is removed`?

test/javascripts/turbolinks_test.coffee
```diff
@@ -122,243 +122,175 @@ describe 'Turbolinks', ->
...
+          assert.equal(Turbolinks.fullPageNavigate.callCount, 1, 'Should not perform a full page refresh.')
           done()
 
-    it 'inserts link with a new href before exiting scripts in head on navigation', (done) ->
-      visit url: 'singleLinkInHead', ->
-        assertLinks(['foo.css'])
-        visit url: 'twoLinksInHeadReverseOrder', ->
-          assertLinks(['bar.css', 'foo.css'])
-          done()
+    describe 'link tags', ->
```
GoodForOneFare: Still don't know why we're sending these events.

test/javascripts/turbolinks_test.coffee
TheMallen: I'm not sure if you are suggesting rewording this or if there is/was another test named that. Looks like there currently isn't.

test/javascripts/turbolinks_test.coffee
TheMallen: Because it fits the conventions of the rest of the lib. We fire `after-node-removed` events and such already. I think going forward we can probably afford to have a conversation about the event lifecycle and what's really needed, I could take or leave these ones for now though.

test/javascripts/turbolinks_test.coffee
TheMallen: `have matching sources` is probably clearest, since it's actually what the lib cares about in this case.

qq99:
:+1:

PR #140, opened 2016-09-06
Fix pushState on page refresh

TheMallen:
This PR makes it so that `pushState` is never called when TG does a full page navigation. This behaviour was causing problems when the server cared about the `referrer` because the destination would essentially be double pushed onto the stack in some situations.

We also introduce tests for this behaviour.

**Reviewers**
@GoodForOneFare 
@qq99

### DIFF
--- a/test/javascripts/turbolinks_test.coffee
+++ b/test/javascripts/turbolinks_test.coffee
@@ -59,6 +59,10 @@ describe 'Turbolinks', ->
       <div id="turbo-area" refresh="turbo-area"></div>
     """
 
+  startFromFixture = (route) ->
+    fixtureHTML = ROUTES[route][2]
+    document.documentElement.innerHTML = fixtureHTML
+
   urlFor = (slug) ->
     window.location.origin + slug
 
@@ -116,6 +120,11 @@ describe 'Turbolinks', ->
         assert.equal('buzz', sandbox.server.requests[0].requestHeaders['fizz'])
         done()
 
+    it 'updates the browser history stack', (done) ->
+      visit url: 'noScriptsOrLinkInHead', ->
+        assert(pushStateStub.called, 'pushState was not called!')
+        done()
+
     it 'calls a user-supplied callback', (done) ->
       yourCallback = stub()
       visit url: 'noScriptsOrLinkInHead', options: {callback: yourCallback}, ->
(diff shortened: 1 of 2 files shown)


qq99:
Nice

test/javascripts/turbolinks_test.coffee
```diff
@@ -135,11 +144,13 @@ describe 'Turbolinks', ->
...
           done()
 
-    describe 'using data-turbolinks-track="true"', ->
-      startFromFixture = (route) ->
-        fixtureHTML = ROUTES[route][2]
-        document.documentElement.innerHTML = fixtureHTML
+    it 'conflicts do not update the browser history stack', (done) ->
+      startFromFixture('singleScriptInHead')
+      visit url: 'singleScriptInHeadWithDifferentSourceButSameName', ->
+        assert(pushStateStub.notCalled, 'pushState was called')
```
GoodForOneFare: `pushState was not called`

test/javascripts/turbolinks_test.coffee
```diff
@@ -135,11 +144,13 @@ describe 'Turbolinks', ->
           assert(Turbolinks.fullPageNavigate.notCalled, 'Should not perform a full page refresh.')
           done()
 
-    describe 'using data-turbolinks-track="true"', ->
-      startFromFixture = (route) ->
-        fixtureHTML = ROUTES[route][2]
-        document.documentElement.innerHTML = fixtureHTML
+    it 'conflicts do not update the browser history stack', (done) ->
```
GoodForOneFare: Nit: this doesn't flow as naturally as other test names.

test/javascripts/turbolinks_test.coffee
```diff
@@ -163,6 +174,12 @@ describe 'Turbolinks', ->
           assert(Turbolinks.fullPageNavigate.called, 'Should perform a full page refresh.')
           done()
 
+      it 'changed assets do not update the browser history stack', (done) ->
```
GoodForOneFare: Nit: this doesn't flow as naturally as other test names.

GoodForOneFare:
Wording aside, LGTM 👍

test/javascripts/turbolinks_test.coffee
TheMallen: when it fails in the console you see

`Turbolinks head asset tracking with turbolinks-track="true" changed assets do not update the browser history stack`

which I think is pretty understandable

test/javascripts/turbolinks_test.coffee
TheMallen: `Turbolinks head asset tracking with conflicts do not update the browser history stack`

hmmmm

maybe it should not be 

`does not update the browser history stack when a conflict is detected`

test/javascripts/turbolinks_test.coffee
TheMallen: Maybe `does not update the browser history stack in cases where it will refresh`

GoodForOneFare: My preference would be to move this out of the anonymous asset tests, and create a `describe` specifically for hash change events.  I'm okay with that happening outside of this PR.

Issue #141, opened 2016-09-06
Replace multiple `visit` calls in tests with `startFromFixture`


TheMallen:
Calling `visit` in tests multiple times is an unfortunate and confusing practice. It makes it hard to stub / spy on things and / or wrap your head around what exactly is going on.

We built `startFromFixture` late in the process of the `0.4.0` work, but didn't roll it out to all of the older tests. We should roll it out ASAP since I've already lost some time to `visit(startState) -> visit(endState) -> asser(foo)` having nonintuitive results a few times.

TheMallen:
Closed by #142

PR #142, opened 2016-09-07
Simplify tests using startFromFixture and minimizing visits

TheMallen:
This PR simplifies tests as suggested in #141.

We minimize multiple calls to `visit` and instead use `startFromFixture` wherever possible. This lets us trust our fixtures to provide the initial state and minimize assertion logic. It should help us avoid silly problems in the future where our tests become broken due to multiple visit calls confusing our stubs / spys. 

note: In some cases we got away with simply removing the first call to `visit` and not bothering with `startFromFixture`.  This is due to the `setup` function already loading an equivalent to `noScriptsOrLinkInHead` every test anyway.

**Reviewers**
@qq99 
@GoodForOneFare


GoodForOneFare:
LGTM 👍 Nice one.

PR #146, opened 2016-09-08
add test for null url to ComponentUrl tests for clarity of intention

TheMallen:
This PR simply adds a test to `ComponentUrl` so that it's behaviour on a null url is clearer to consumers.

@Shopify/tnt 
@qq99 
@GoodForOneFare

### DIFF
--- a/test/javascripts/component_url_test.coffee
+++ b/test/javascripts/component_url_test.coffee
@@ -1,6 +1,6 @@
 describe 'ComponentUrl', ->
   describe 'constructor', ->
-    it 'uses current location when given not given a url', ->
+    it 'uses current location when not given a url', ->
       url = new ComponentUrl()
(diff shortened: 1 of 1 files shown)


test/javascripts/component_url_test.coffee
```diff
@@ -1,5 +1,9 @@
 describe 'ComponentUrl', ->
   describe 'constructor', ->
+    it 'uses current location when given not given a url', ->
```
arapehl: >given not given
?

qq99:
👍

test/javascripts/component_url_test.coffee
TheMallen: Oops/

GoodForOneFare:
👍 🚢 🐑

Issue #151, opened 2016-09-15
Turbohead may insert duplicate scripts and/or execute inline scripts early


GoodForOneFare:
With a slow connection, it's easy to fool Turbohead into doing pathological things :/  This may be two separate issues; feel free to split them up.

-----

**Environment setup**
* Point your Gemfile at a local version of Turbograft and `dev up`
* Alter `Turbohead#insertAssetTask` to include this debug statement:
```coffeescript
insertAssetTask = (activeDocument, newNode, name) ->
  (done) ->
    console.log("Turbohead::insertAssetTask - inserting", newNode.src)
```
* In Chrome dev tools to use _Disabled cache_ & _Regular 3G_

-----

**Reproducing the duplicate scripts bug**
* Open your dev console
* Go to Home
* Refresh the browser
* Quickly navigate to _Products_
* Wait for the `insertAssetTask` message to appear for `admin-vendors-others`
* Quickly navigate to _Transfers_

![shop 1 transfers shopify 2016-09-15 01-18-16](https://cloud.githubusercontent.com/assets/673655/18539109/978f9efe-7ae2-11e6-9324-bc07f22b7049.png)

-----

**Reproducing the early inline scripts bug**
* Open your dev console
* Go to Home
* Refresh the browser
* Quickly navigate to _Products_
* Wait for the `insertAssetTask - inserting index-0123456789.js` message to appear
* Quickly navigate to _Orders_
* Note that inline scripts are executed as soon as _Orders_ HTML is received (before `index.js` has finished downloading)

![shop 1 orders shopify 2016-09-15 01-12-04](https://cloud.githubusercontent.com/assets/673655/18539018/caaf3cb4-7ae1-11e6-99d3-f0b544bc047b.png)

justinthec:
😢 the inner machinations of turbograft are an enigma

GoodForOneFare:
@justinthec do you ever sleep?  Patches welcome... :P

edward:
> Point your Gemfile at a local version of Turbograft and dev up

Note that you can also do `bundle open turbograft` and your editor will open up the installed gem source for you to muck about in. You might have to set `$EDITOR` or `$BUNDLER_EDITOR` like so: `EDITOR=atom bundle open turbograft` or one of the many other ways to set environment variables in your shell.

PR #152, opened 2016-09-16
Fix concurrent requests

GoodForOneFare:
Works around a couple of bugs caused by users navigating around quickly (see #151 for reproduction scenarios).  Basically, this change ensures that JS assets are _definitely_ loaded before trying to execute inline scripts.

/cc @TheMallen, @qq99

### DIFF
--- a/lib/assets/javascripts/turbograft/turbohead.coffee
+++ b/lib/assets/javascripts/turbograft/turbohead.coffee
@@ -3,7 +3,7 @@ TRACKED_ATTRIBUTE_NAME = 'turbolinksTrack'
 ANONYMOUS_TRACK_VALUE = 'true'
 
 scriptPromises = {}
-rejectPreviousRequest = null
+resolvePreviousRequest = null
 
 waitForCompleteDownloads = ->
   loadingPromises = Object.keys(scriptPromises).map (url) ->
@@ -24,7 +24,7 @@ class window.TurboHead
 
   @reset: ->
     scriptPromises = {}
-    rejectPreviousRequest = null
+    resolvePreviousRequest = null
 
   hasChangedAnonymousAssets: () ->
     anonymousUpstreamAssets = @upstreamAssets
@@ -52,10 +52,10 @@ class window.TurboHead
     @hasNamedAssetConflicts() || @hasChangedAnonymousAssets()
 
   waitForAssets: () ->
-    rejectPreviousRequest?()
+    resolvePreviousRequest?(isCanceled: true)
 
-    new Promise((resolve, reject) =>
-      rejectPreviousRequest = reject
+    new Promise((resolve) =>
+      resolvePreviousRequest = resolve
       waitForCompleteDownloads()
         .then(@_insertNewAssets)
         .then(waitForCompleteDownloads)
--- a/lib/assets/javascripts/turbograft/turbohead.coffee
+++ b/lib/assets/javascripts/turbograft/turbohead.coffee
@@ -2,6 +2,14 @@ TRACKED_ASSET_SELECTOR = '[data-turbolinks-track]'
 TRACKED_ATTRIBUTE_NAME = 'turbolinksTrack'
 ANONYMOUS_TRACK_VALUE = 'true'
 
+scriptPromises = {}
+rejectPreviousRequest = null
+
+waitForCompleteDownloads = ->
+  loadingPromises = Object.keys(scriptPromises).map (url) ->
+    scriptPromises[url]
+  Promise.all(loadingPromises)
+
 class window.TurboHead
   constructor: (@activeDocument, @upstreamDocument) ->
     @activeAssets = extractTrackedAssets(@activeDocument)
@@ -13,6 +21,10 @@ class window.TurboHead
     @newLinks = @upstreamAssets
       .filter(attributeMatches('nodeName', 'LINK'))
       .filter(noAttributeMatchesIn('href', @activeAssets))
+
+  @reset: ->
+    scriptPromises = {}
+    rejectPreviousRequest = null
 
   hasChangedAnonymousAssets: () ->
     anonymousUpstreamAssets = @upstreamAssets
@@ -39,9 +51,20 @@ class window.TurboHead
   hasAssetConflicts: () ->
     @hasNamedAssetConflicts() || @hasChangedAnonymousAssets()
 
-  insertNewAssets: (callback) ->
+  waitForAssets: () ->
+    rejectPreviousRequest?()
+
+    new Promise((resolve, reject) =>
+      rejectPreviousRequest = reject
+      waitForCompleteDownloads()
+        .then(@_insertNewAssets)
+        .then(waitForCompleteDownloads)
+        .then(resolve)
+    )
+
+  _insertNewAssets: () =>
     updateLinkTags(@activeDocument, @newLinks)
-    updateScriptTags(@activeDocument, @newScripts, callback)
+    updateScriptTags(@activeDocument, @newScripts)
 
 extractTrackedAssets = (doc) ->
   [].slice.call(doc.querySelectorAll(TRACKED_ASSET_SELECTOR))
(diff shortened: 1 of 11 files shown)


lib/assets/javascripts/turbograft/turbohead.coffee
```diff
@@ -14,6 +22,10 @@ class window.TurboHead
       .filter(attributeMatches('nodeName', 'LINK'))
       .filter(noAttributeMatchesIn('href', @activeAssets))
 
+  @reset: ->
```
GoodForOneFare: I dislike exposing a test-only method, but the alternative is assigning statics vars to the class.  Either way, we're exposing more context than I want to :/

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -139,7 +139,11 @@ class window.Turbolinks
         if turbohead.hasAssetConflicts()
           return Turbolinks.fullPageNavigate(url.absolute)
         reflectNewUrl url if options.updatePushState
-        turbohead.insertNewAssets(-> updateBody(upstreamDocument, xhr, options))
+        turbohead.waitForAssets().then(->
+          updateBody(upstreamDocument, xhr, options)
+        ).catch(->
+          # A new request appeared during asset download.
```
GoodForOneFare: Is this abusing promises?  Does a request being superseded really count as a rejection?

An alternative to catching would be resolving with an `isCancelled` flag, and only calling `updateBody` if that's `false`.

test/javascripts/vendor/promise-polyfill/promise.js
```diff
@@ -0,0 +1,233 @@
+(function (root) {
```
GoodForOneFare: A future commit will pull this in via NPM (if we agree that's a good idea).

test/javascripts/test_helper.js
```diff
@@ -11,3 +12,5 @@ mock = sinon.mock;
 stub = sinon.stub;
 
 mocha.setup('tdd')
+sinon.assert.expose(chai.assert, {prefix: ''});
```
GoodForOneFare: This allows stuff like `assert.lengthOf`.

test/javascripts/remote_test.coffee
```diff
@@ -94,10 +94,7 @@ describe 'Remote', ->
         fail: done
       , @initiating_target
 
-      simulateError = ->
-        listener(target: remote.xhr) for listener in remote.xhr.eventListeners.error
-
-      setTimeout(simulateError, 0)
+      remote.xhr.respond(404)
```
GoodForOneFare: The previous version of this test spit out some distracting error messages (an NPE on xhr).

lib/assets/javascripts/turbograft/turbohead.coffee
TheMallen: Maybe we have the domain model wrong?

Maybe a `Turbohead` instance wants to be where the cache lives, created at time of construction, and then a `process` method on a turbohead object returns some kind of `headAssets` object which we use for `update`. Then we'd have `Turbolinks` create it's `Turbohead`, and in tests you could just create a new one when you need to.

lib/assets/javascripts/turbograft/turbolinks.coffee
TheMallen: I think that makes more sense personally.

lib/assets/javascripts/turbograft/turbohead.coffee
```diff
@@ -76,37 +99,37 @@ noDatasetMatchesIn = (attribute, collection) ->
...
+    triggerEvent("page:after-link-inserted", newNode)
   )
 
-asyncSeries = (tasks, callback) ->
-  return callback() if tasks.length == 0
-  task = tasks.shift()
-  task(-> asyncSeries(tasks, callback))
+updateScriptTags = (activeDocument, newScripts) ->
+  promise = Promise.resolve()
+  newScripts.forEach (scriptNode) ->
```
TheMallen: This can be made much cleaner with `Promise.all`, since it will be the same as `Promise.resolve` when called on an empty array.

```
scriptInsertions = newScripts.map(-> insertScript(activeDocument, scriptNode))
Promise.all(scriptInsertions)
```
or even
```
Promise.all(newScripts.map(-> insertScript(activeDocument, scriptNode)))
```
if you're feeling one linery.

test/javascripts/vendor/promise-polyfill/promise.js
TheMallen: ¯\_(ツ)_/¯

Personally I think npm is the Right Way to get javascript into your app. I can understand not wanting to add the complexity for a small polyfill only used in tests, however.

lib/assets/javascripts/turbograft/turbohead.coffee
GoodForOneFare: `Promise.all` will execute them in parallel.  That's the opposite of what we want here.

TheMallen: Right. : /

lib/assets/javascripts/turbograft/turbolinks.coffee
GoodForOneFare: Pushed up an initial implementation of this.  Still trying to make a `turbolinks does not update body when request is superseded` work :/

test/javascripts/vendor/promise-polyfill/promise.js
GoodForOneFare: The rationale behind not using NPM was:
- This is a critical fix that I was trying to get out in a day
- You said something about NPM imports not always working when I asked about using NPM

In an ideal world, we'd be using NPM to import promise + modern versions sinon, chai, etc.

lib/assets/javascripts/turbograft/turbohead.coffee
GoodForOneFare: Yes, agree that the domain model needs to be adjusted to the new requirements.  As we discussed in Slack, this isn't a refactoring PR.

GoodForOneFare:
Instead of rejecting promises, a cancelled request now resolves with an `isCanceled` flag.  Added tests for that to `turbolinks_test` and `turbohead_test`.

@TheMallen can I get a 🎩 and follow-up review, please
@qq99 this is fixing an error that's affecting the user experience; will you have time to review this today?

GoodForOneFare:
Looks like Anthony is hella busy today.  @Shopify/tnt can someone take a look at this, please?

TheMallen:
@GoodForOneFare Checking this out

lib/assets/javascripts/turbograft/turbohead.coffee
```diff
@@ -39,9 +51,20 @@ class window.TurboHead
   hasAssetConflicts: () ->
     @hasNamedAssetConflicts() || @hasChangedAnonymousAssets()
 
-  insertNewAssets: (callback) ->
+  waitForAssets: () ->
+    resolvePreviousRequest?(isCanceled: true)
```
TheMallen: Am I crazy or is the spelling usually `cancelled`?

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -139,7 +139,9 @@ class window.Turbolinks
         if turbohead.hasAssetConflicts()
           return Turbolinks.fullPageNavigate(url.absolute)
         reflectNewUrl url if options.updatePushState
-        turbohead.insertNewAssets(-> updateBody(upstreamDocument, xhr, options))
+        turbohead.waitForAssets().then((result) ->
+          updateBody(upstreamDocument, xhr, options) unless result?.isCanceled
```
TheMallen: 💯

lib/assets/javascripts/turbograft/turbohead.coffee
GoodForOneFare: US English has one l.  Looks funky to me, but then so does "color".

TheMallen:
Finally starting 🔝 👒 , will let you know how it goes

lib/assets/javascripts/turbograft/turbohead.coffee
DrewMartin: We could hide it inside of an object called `_testAPI` or something. We use that in the theme editor and I like it

TheMallen:
## Fix Results
Throttled connection, waited for these to appear, rapidly switched to several other pages, final page loaded fine and only triggered the two expected inserts.

image


## General Behaviour
- all the usual stuff worked 
  - expired session
  - random CRUD
  - theme editor redirects
  - etc.

lib/assets/javascripts/turbograft/turbohead.coffee
GoodForOneFare: Done!  Thanks!

lib/assets/javascripts/turbograft/turbohead.coffee
```diff
@@ -76,37 +101,37 @@ noDatasetMatchesIn = (attribute, collection) ->
...
+  url = scriptNode.src
+  if scriptPromises[url]
+    return scriptPromises[url]
 
-insertScriptTask = (activeDocument, scriptNode) ->
-  # We need to clone script tags in order to ensure that the browser executes them.
+  # Clone script tags to guarantee browser execution.
   newNode = activeDocument.createElement('SCRIPT')
   newNode.setAttribute(attr.name, attr.value) for attr in scriptNode.attributes
   newNode.appendChild(activeDocument.createTextNode(scriptNode.innerHTML))
```
DrewMartin: I'm unclear how this part works. We clone the script tags here, but then also again in [turbolinks.coffee](https://github.com/Shopify/turbograft/blob/b639629df256fcd1045b6665af75ec4d5fffd9fa/lib/assets/javascripts/turbograft/turbolinks.coffee#L285-L292)? In this case the second clone only happens if `isCanceled` is true, but won't that mean they get executed once or twice, rather than never or once?

GoodForOneFare: The cloned script tags in `turbolinks` are exclusively inline scripts from the new `document.body`.  `turbohead` is cloning tags from `document.head`.

DrewMartin: Ohhhh ok

lib/assets/javascripts/turbograft/turbohead.coffee
```diff
@@ -76,37 +101,37 @@ noDatasetMatchesIn = (attribute, collection) ->
...
 
-insertLinkTask = (activeDocument, node) ->
-  insertAssetTask(activeDocument, node.cloneNode(), 'link')
-
-insertAssetTask = (activeDocument, newNode, name) ->
-  (done) ->
+  scriptPromises[url] = new Promise((resolve) ->
     onAssetEvent = (event) ->
-      triggerEvent("page:#{name}-error", event) if event.type == 'error'
+      triggerEvent("page:#script-error", event) if event.type == 'error'
```
DrewMartin: I guess it's not really a change in behaviour. This would be a separate thing if it's even an issue

DrewMartin:
👍

lib/assets/javascripts/turbograft/turbohead.coffee
GoodForOneFare: Hmm, that's an excellent point.  I *think* any scripts can be removed from `scriptPromises` at this point.  I'll let that soak overnight and make the change if I can't think of a problematic scenario.

To be clear, though, the error script will still be in `document.head`, so it'll never be reloaded.

GoodForOneFare:
RE cleaning out `scriptPromises`.  Because we calculate `newScripts` as soon as a `Turbohead` instance is created, but then use it (potentially seconds later) in `_insertNewAssets`, we're relying on `scriptPromises` to guard against tags that have been inserted by another request.  

Clearing out scripts that result in `error` events *may* get them to rerun, but not consistently.

I'm going to merge this as is because we want to stem the flow of user errors in Bugsnag.  But I'll think it over... at the very least, I need to add a test case to prove what I've just said is true.

PR #154, opened 2016-09-22
Fix case of moving from a layout with tracked assets to one with none

### LINKED ISSUE #147: Navigating from page with named assets to page with none should refresh
## Bug
When you navigate from a page that has named tracked assets to a page with none you end up keeping all the styles from the page that had them. In our case this means the login page ends up looking broken when a user's session expires unless we follow the workaround. (only in the bundle split branch)

Version:  `v0.4.1`
Workaround: Put a `data-turbolinks-track="true"` on any `link` or `head` asset in the page with no named assets. TG will see that the set of anonymous tracked elements has gone from 0 to 1 and cause a refresh.

## Proposed Solution
When TG goes from a page with tracked assets of any kind to a page with 0 tracked assets of any kind, it should force a refresh. I can think of no situations where you might actively want the current behaviour that aren't really bad ideas.

TheMallen:
This PR fixes #147

We'll be able to drop the workaround.

@Shopify/tnt 
@edward
@GoodForOneFare 
@utkarshsaxenashopify

### DIFF
--- a/lib/assets/javascripts/turbograft/turbohead.coffee
+++ b/lib/assets/javascripts/turbograft/turbohead.coffee
@@ -44,6 +44,9 @@ class window.TurboHead
       noMatchingSrc(node) || noMatchingHref(node)
     )
 
+  movingFromTrackedToUntracked: () ->
+    @upstreamAssets.length == 0 && @activeAssets.length > 0
+
   hasNamedAssetConflicts: () ->
     @newScripts
       .concat(@newLinks)
@@ -51,7 +54,9 @@ class window.TurboHead
       .some(datasetMatchesIn(TRACKED_ATTRIBUTE_NAME, @activeAssets))
 
   hasAssetConflicts: () ->
-    @hasNamedAssetConflicts() || @hasChangedAnonymousAssets()
+    @movingFromTrackedToUntracked() ||
+      @hasNamedAssetConflicts() ||
+      @hasChangedAnonymousAssets()
 
   waitForAssets: () ->
     resolvePreviousRequest?(isCanceled: true)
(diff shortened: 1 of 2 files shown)


TheMallen:
🎩 instructions:

- Go to `layouts/auth.html.erb` in `Shopify/Shopify`
- Remove any `data-turbolinks-track="true"` from the head (this was the workaround)
- Boot up your server
- Log in
- Expire your session,
- Navigate anywhere
-  styles should be wrecked on the login screen
- Clone this branch
- Go to `gemfile` point it to the local copy
- Boot up your server again
- Log in
- Expire your session
- Navigate anywhere
-  styles should be 👌

GoodForOneFare:
While this works, it seems like a half fix.  It covers `refresh when navigating to a page with no assets`, but I don't think it's going to handle `refresh when navigating to a TG page with completely different assets`.  Is there a solution that covers both?

TheMallen:
It covers refresh when navigating to a page with no *tracked* assets. We could check the cross section of head assets length and then that would give us a refresh in this case as a subset of the no overlap case. I'm not certain we should refresh when you head to a page with no crossover however. I think we cross that bridge when we come to it.

GoodForOneFare:
I'm not certain, either.  That's why we should formalize something to make this explicit.

I mean, 👍 LGTM on this scrappy thing being better than another scrappy thing, but I think we should ultimately replace this with something less magical.

TheMallen:
Based on IRL discussion we decided we can defer making this more generic until we have some idea of the use cases that need it, since we have several possible solution paths but no context to know which is the right abstraction, and don't think it's a good use of time atm.

qq99:
This PR sounds reasonable 👍

PR #144, opened 2016-09-08
Refactor Process Doc, Add new Error

TheMallen:
This PR accomplishes two primary goals:
- Address the occasional spikes of `Cannot read property 'match' of null` we've been getting for ages, and hopefully adds enough information to those cases that we can figure out why they're happening.
- Iteratively add some testing / refactor some code out of the main Turbolinks file.

@qq99 
@GoodForOneFare

Issue #138, opened 2016-09-02
Turbolinks should accept a document parameter


TheMallen:
Turbolinks should allow you to set a document for it to operate on. This would allow it to control iframes and the like, or possibly do it's manipulations and then have something like morphdom perform a diff afterwards, as well as allow tests to be done using a document fragment, thus making Teaspoon's test runner more useful.

Issue #86, opened 2015-04-16
Try to move test runner to PhantomJS 2.0


qq99:
Perhaps a handy blog post:
https://mediocre.com/forum/topics/phantomjs-2-and-travis-ci-we-beat-our-heads-against-a-wall-so-you-dont-have-to

TheMallen:
Done :)

PR #155, opened 2016-09-22
Fix/scroll reset

### LINKED ISSUE #149: tg-remote should reset scrolling when it receives a redirect response
**Example**
* Tweak cards/apple_pay_supported_card.rb#qualifies? to always return true
* Open dev tools and view the Network tab
* View your home feed
* Scroll down to the card, *wait 5 seconds* and click the Activate Apple Pay button
* Note that the document body is still scrolled, when it should be at the top of the payments section

TheMallen:
## Summary
This PR fixes #149 and does some minor refactoring. Notably it simplifies the confusing dynamic between `onLoadFunction` and `callback` and moves `option` finagling closer to where the props are used (out of `fetch`). 

## Testing
Unfortunately, writing an automated test for this will not be possible until we make some changes to the test setup. This problem is related to #138, and #86, as we can't currently actually check scroll positions meaningfully in tests with our current setup.  ¯\_(ツ)_/¯

To 🎩  that this fixes the bug, click any card on home that triggers a redirect and observe the beautiful scroll reset.

You'll also want to at minimum make sure that:
- partial replaces do not reset scroll (play around on products or to make sure of this)
- full visits do reset scroll (scroll down a page and then click somewhere in the nav bar)
- tg-remotes do not reset scroll (pretty much any form in the admin)

**note: I want to significantly refactor the implementation of this fix after #144 merges, moving towards having the logic that cares about the xhr's data inside of the new `Request` object**

CC @Shopify/tnt 
@edward 
@GoodForOneFare 
@utkarshsaxenashopify
@qq99

### DIFF
--- a/lib/assets/javascripts/turbograft/page.coffee
+++ b/lib/assets/javascripts/turbograft/page.coffee
@@ -15,18 +15,15 @@ Page.refresh = (options = {}, callback) ->
     location.pathname + paramString
   else
     location.href
+  delete options.url
 
+  options.partialReplace = true
+  options.callback = callback if callback
   if options.response
-    options.partialReplace = true
-    options.onLoadFunction = callback
-
     xhr = options.response
     delete options.response
     Turbolinks.loadPage null, xhr, options
   else
-    options.partialReplace = true
-    options.callback = callback if callback
-
     Turbolinks.visit newUrl, options
 
 Page.open = ->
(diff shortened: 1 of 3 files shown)


TheMallen:
if CI goes green on the rebased + squashed PR this is ready to go (will refactor in a subsequent PR)

TheMallen:
This is sadly, broken. The fix assumed that we wanted to resetScroll on redirects. This is actually not the case, because most redirects are the result of regular form actions where we are partial replacing in the updated form and don't want to lose ui state. In fact we only want to reset scroll in very specific circumstances. I'm closing this for now.

PR #156, opened 2016-09-23
[WIP] allow tg-remote to control history state

qq99:
WIP

### DIFF
--- a/lib/assets/javascripts/turbograft/remote.coffee
+++ b/lib/assets/javascripts/turbograft/remote.coffee
@@ -12,6 +12,7 @@ class TurboGraft.Remote
     @refreshOnSuccessExcept = @opts.refreshOnSuccessExcept.split(" ") if @opts.refreshOnSuccessExcept
     @refreshOnError         = @opts.refreshOnError.split(" ")         if @opts.refreshOnError
     @refreshOnErrorExcept   = @opts.refreshOnErrorExcept.split(" ")   if @opts.refreshOnErrorExcept
+    @updatePushState        = @opts.updatePushState
 
     xhr = new XMLHttpRequest
     if @actualRequestType == 'GET'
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -12,6 +12,7 @@ class TurboGraft.Remote
     @refreshOnSuccessExcept = @opts.refreshOnSuccessExcept.split(" ") if @opts.refreshOnSuccessExcept
     @refreshOnError         = @opts.refreshOnError.split(" ")         if @opts.refreshOnError
     @refreshOnErrorExcept   = @opts.refreshOnErrorExcept.split(" ")   if @opts.refreshOnErrorExcept
+    @updatePushState        = @opts.updatePushState
```
qq99: maybe should `|| true` here

TheMallen: I'm assuming you mean if they ever use the constructor rather than letting an initializer do it. Currently in admin we have `Shopify.Remote` which is often instantiated directly from within a class. Assuming we want to move those to `Turbograft.Remote` then we already have some places that would do this.

lib/assets/javascripts/turbograft/remote.coffee
```diff
@@ -130,20 +131,26 @@ class TurboGraft.Remote
 
     unless TurboGraft.hasTGAttribute(@initiator, 'tg-remote-norefresh')
       if @opts.fullRefresh && @refreshOnSuccess
-        Page.refresh(onlyKeys: @refreshOnSuccess)
+        Page.refresh
+          onlyKeys: @refreshOnSuccess
+          updatePushState: @updatePushState
```
TheMallen: I don't like manually passing these in every call. I had to do it in #162 as well, so I went with:
image

It might be even cleaner to just make a function that wraps `Page.refresh` and always forwards the arguments we want instead of calling it using the result of another function though.

PR #158, opened 2016-09-23
add a check for redirection before setting options.partialReplace

### LINKED ISSUE #157: Remote redirects should wait for head scripts
If a page uses `<a href="/foo" data-tg="remote">`, and `/foo` results in a redirect, Turbograft should wait for any new `<head>` assets to load.

See https://github.com/Shopify/shopify/pull/85634 for more details.

/cc @TheMallen

TheMallen:
This PR fixes #157.

When a remote action causes`Page.refresh` to be given a redirect, we now no longer mark it as a partial replace. This means that we gain head asset tracking and stop breaking stuff. Wooh.

This could have been a one line change but I wanted to make it more obvious why I was doing the check, hence the variable.

@Shopify/tnt 
@GoodForOneFare 
@qq99 
@edward

### DIFF
--- a/lib/assets/javascripts/turbograft/page.coffee
+++ b/lib/assets/javascripts/turbograft/page.coffee
@@ -17,7 +17,9 @@ Page.refresh = (options = {}, callback) ->
     location.href
 
   if options.response
-    options.partialReplace = true
+    followingARedirect = options.response.responseURL != newUrl
+    options.partialReplace = true unless followingARedirect
+
     options.onLoadFunction = callback
 
     xhr = options.response


GoodForOneFare:
LGTM 👍

TheMallen:
Yes, I'll add tests for it. First got to figure out why all the tests are suddenly breaking though.

TheMallen:
I got stuck on some silliness, I'll fix this tommorrow :)

TheMallen:
This is also broken for similar reasons to the other PR. I'll open one that addresses redirection logic more thoroughly in a subsequent PR.

PR #160, opened 2016-09-27
fix xhr header application

### LINKED ISSUE #148: X-XHR-Redirected-To is sometimes applied to the wrong response
**Context**
[xhr_headers#store_for_turbolinks](https://github.com/turbolinks/turbolinks-classic/blob/master/lib/turbolinks/xhr_headers.rb#L26) stores redirect info in `session[:_turbolinks_redirect_to]`.  [set_xhr_redirected_to](https://github.com/turbolinks/turbolinks-classic/blob/master/lib/turbolinks/xhr_headers.rb#L32) fishes out this information, and applies it to a response.

**Problem**
If the browser makes multiple requests during an operation, the redirect info is applied to the first processed request.

**Example**
* Tweak `cards/apple_pay_supported_card.rb#qualifies?` to always return `true`
* Open dev tools and view the _Network_ tab
* View your home feed
* Scroll down to the card, and **quickly** click the _Activate Apple Pay_ button
* Note that the browser sends up a few requests:
  * _admin/home/consume_cards_
  * _admin/home/register_card_enagement_
  * _admin/settings/apple_pay_activation_
* Click through the response headers and note that `X-XHR-Redirected-To` is assigned to whichever request returns first

/cc @TheMallen

TheMallen:
This PR fixes #148.

~~I'm not sure how to write tests to show this fixes the issue (my ruby weakness is showing),~~ but I was able to test it in all the major browsers using pry and going through the steps mentioned in the issue.

It does however raise issues with the fixes in one of my other open PRs, so I'll have to address those once this is merged in.

@Shopify/tnt 
@GoodForOneFare 
@qq99

### DIFF
--- a/lib/turbograft/xhr_headers.rb
+++ b/lib/turbograft/xhr_headers.rb
@@ -25,23 +25,32 @@ module TurboGraft
     end
 
     private
-      def store_for_turbolinks(url)
-        session[:_turbolinks_redirect_to] = url if session && request.headers["X-XHR-Referer"]
-        url
-      end
 
-      def set_xhr_redirected_to
-        if session && session[:_turbolinks_redirect_to]
-          response.headers['X-XHR-Redirected-To'] = session.delete :_turbolinks_redirect_to
-        end
-      end
+    # Ensure backwards compatibility
+    # Rails < 4.2:  _compute_redirect_to_location(options)
+    # Rails >= 4.2: _compute_redirect_to_location(request, options)
+    def _normalize_redirect_params(args)
+      options, req = args.reverse
+      [options, req || request]
+    end
 
-      # Ensure backwards compatibility
-      # Rails < 4.2:  _compute_redirect_to_location(options)
-      # Rails >= 4.2: _compute_redirect_to_location(request, options)
-      def _normalize_redirect_params(args)
-        options, req = args.reverse
-        [options, req || request]
+    def store_for_turbolinks(url)
+      session[:_turbolinks_redirect_to] = url if session && request.headers["X-XHR-Referer"]
+      url
+    end
+
+    def set_xhr_redirected_to
+      if turbolinks_redirected? && request_matches_redirect?
+        response.headers['X-XHR-Redirected-To'] = session.delete(:_turbolinks_redirect_to)
       end
+    end
+
+    def turbolinks_redirected?
+      session && session[:_turbolinks_redirect_to]
+    end
+
+    def request_matches_redirect?
+      session[:_turbolinks_redirect_to] == request.original_url
+    end
   end
 end
--- a/test/controller/legacy_pages_controller_test.rb
+++ b/test/controller/legacy_pages_controller_test.rb
@@ -37,18 +37,15 @@ class LegacyPagesControllerTest < ActionController::TestCase
     assert_nil session[:_turbolinks_redirect_to]
   end
 
-  test "set_xhr_redirected_to sets X-XHR-Redirected-To" do
-    get :index, {}, {_turbolinks_redirect_to: 'http://test.host/expected'}
-    assert_response :ok
-    assert_equal 'http://test.host/expected', response.headers['X-XHR-Redirected-To']
+  test "XHR GET for the page matching the redirect will set X-XHR-Redirected-To" do
+    session[:_turbolinks_redirect_to] = 'http://test.host/legacy_pages/321'
+    get :show, params: {id: 321}, xhr: true
+    assert_equal 'http://test.host/legacy_pages/321', response.headers['X-XHR-Redirected-To']
   end
 
-  test "XHR POST to a redirecting route, followed by XHR GET will set X-XHR-Redirected-To" do
-    @request.headers["X-XHR-Referer"] = 'http://test.host'
-    xhr :post, :redirect_to_somewhere_else_after_POST
-    assert_response 302
-    assert_redirected_to legacy_page_path(321)
-    xhr :get, :show, id: 321
-    assert_equal 'http://test.host/legacy_pages/321', response.headers['X-XHR-Redirected-To']
+  test "XHR GET for a page not matching the redirect will not set X-XHR-Redirected-To" do
+    session[:_turbolinks_redirect_to] = 'http://test.host/legacy_pages/321'
+    get :show, params: {id: 11}, xhr: true
+    assert_nil response.headers['X-XHR-Redirected-To']
   end
 end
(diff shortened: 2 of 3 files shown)


TheMallen:
Guess I'll have to fix these tests.

TheMallen:
Update: the tests are fixed now afaik. I encountered a lot of wierdness, which I will attempt to explain.

image
 
So to recap, the issue this PR is fixing is that our headers would be applied to *whatever the next request was* instead of to the matching request for the redirect. In this case, everything looks like it should be 👌 , but it turns out that something railsy is happening behind the scenes that *forces the url of the second request to be pointed at the same place as the original one*. You can solve this by emptying the request, but then we lose the session. Seeing as that means we'd be setting it manually anyway, it made more sense to just test that we 

A) set the header correctly
B) consume the header from session when the url matches
C) **don't** consume the header from the session when the url doesn't match

Since we already had a test for setting the header, I just deleted the broken tests and replaced them with the B and C above.

It's possible that my lack of ruby/rails expertise is making me miss something obvious here. Please let me know if that is the case.

TheMallen:
Looks like the legacy test is still borked. :/

lib/turbograft/xhr_headers.rb
```diff
@@ -31,7 +31,7 @@ def store_for_turbolinks(url)
       end
 
       def set_xhr_redirected_to
-        if session && session[:_turbolinks_redirect_to]
+        if session && session[:_turbolinks_redirect_to] && session[:_turbolinks_redirect_to] == request.original_url
```
edward: How do you feel about adding a clarifying method like this?

```ruby
def original_request?
  session[:_turbolinks_redirect_to] == request.original_url
end
```

edward: Or maybe `first_original_request?`

lib/turbograft/xhr_headers.rb
```diff
@@ -31,7 +31,7 @@ def store_for_turbolinks(url)
...
           response.headers['X-XHR-Redirected-To'] = session.delete :_turbolinks_redirect_to
```
edward: `session.delete(:_turbolinks_redirect_to)`

test/controller/legacy_pages_controller_test.rb
```diff
@@ -37,18 +37,15 @@ class LegacyPagesControllerTest < ActionController::TestCase
     assert_nil session[:_turbolinks_redirect_to]
   end
 
-  test "set_xhr_redirected_to sets X-XHR-Redirected-To" do
-    get :index, {}, {_turbolinks_redirect_to: 'http://test.host/expected'}
-    assert_response :ok
-    assert_equal 'http://test.host/expected', response.headers['X-XHR-Redirected-To']
+  test "XHR get for the page matching the redirect will set X-XHR-Redirected-To" do
```
GoodForOneFare: I'm going to need a crash course in Railsy testing to get what's going on here.

TheMallen: If you want to go through it in person, I'm down.

lib/turbograft/xhr_headers.rb
TheMallen: Sure

TheMallen: Though I think that `original_request` is probably misleading, we're looking for the request that is pointed where we told the first one to redirect to. It would be more like

`request_matches_redirect?`

test/controller/pages_controller_test.rb
```diff
@@ -37,18 +37,15 @@ class PagesControllerTest < ActionController::TestCase
...
   end
 
-  test "XHR POST to a redirecting route, followed by XHR GET will set X-XHR-Redirected-To" do
-    @request.headers["X-XHR-Referer"] = 'http://test.host'
-    xhr :post, :redirect_to_somewhere_else_after_POST
-    assert_response 302
-    assert_redirected_to page_path(321)
-    xhr :get, :show, id: 321
-    assert_equal 'http://test.host/pages/321', response.headers['X-XHR-Redirected-To']
+  test "XHR get for a page not matching the redirect will not set X-XHR-Redirected-To" do
```
edward: Small thing but I find it easier to parse/recognize the distinction between a HTTP `GET` operation and normal English “get” when the HTTP version of the word is in all-caps.

lib/turbograft/xhr_headers.rb
```diff
@@ -25,23 +25,31 @@ def _compute_redirect_to_location(*args)
...
+    end
 
-      # Ensure backwards compatibility
-      # Rails < 4.2:  _compute_redirect_to_location(options)
-      # Rails >= 4.2: _compute_redirect_to_location(request, options)
-      def _normalize_redirect_params(args)
-        options, req = args.reverse
-        [options, req || request]
+    def set_xhr_redirected_to
+      if redirect_was_sent? && request_matches_redirect?
```
edward: I was a little confused by whether the redirect was being sent or received – what about renaming it to `redirected?`, `xhr_redirected?`, or `turbolinks_redirected?`

test/controller/legacy_pages_controller_test.rb
```diff
@@ -37,18 +37,15 @@ class LegacyPagesControllerTest < ActionController::TestCase
...
+    session[:_turbolinks_redirect_to] = 'http://test.host/legacy_pages/321'
+    get :show, params: {id: 321}, xhr: true
```
edward: You can set the session in this line as well if you think it would be more clear:

http://api.rubyonrails.org/v4.2.7.1/classes/ActionController/TestCase/Behavior.html#method-i-get
image

test/controller/pages_controller_test.rb
TheMallen: 🆗

lib/turbograft/xhr_headers.rb
TheMallen: `turbolinks_redirected?` seems good

test/controller/legacy_pages_controller_test.rb
TheMallen: I kind of like it as is personally

GoodForOneFare:
👍 LGTM.  Thanks for guiding me through the tests.  They make sense today (brain fog sucks).

GoodForOneFare:
Note that I didn't 🎩.  Let me know if that would be useful.

TheMallen:
You could functional test if you want, as in, you can throttle your connection and do something that triggers multiple XHRs (one of which redirects) and check that the right one ends up with the header. That's a bit excessive though ;)

TheMallen:
@qq99 
@DrewMartin 
Can you guys take a look at this when you have a chance?

TheMallen:
ALRIGHTY, I'd like to merge this. Got 2 OKs, just gonna wait for CI on the squashed version.

PR #162, opened 2016-09-28
Fix remote redirects being all wonky

TheMallen:
## Summary
This PR fixes the two main issues (#157 and #149) with remotes that result in redirects ~~on an opt-in  basis~~, and also refactors some of `Turbolinks` to make it a bit easier to maintain (and to add these changes). 

## Details
As we can see with #157 and #149, remotes that result in redirects can be pretty sketchy in the current release of TG. In most cases, a remote resulting in a redirect is due to submitting a form and then being redirected to the html generating endpoint afterwards. This means that we want scroll position to not be affected, and we don't want to bother waiting on head asset tracking. However, some cases (see home card navigation) actually are page navigations in disguise. In these (exceptional) cases there is no elegant solution with the current build of TG and we are currently just forcing full page refreshes on them. 

## How?
- Adds a check for whether the `X-XHR-Redirect` header is pointing elsewhere, and considers the visit to be a non-partial replace if so. (This causes a scroll reset and triggers turbohead tracking)
- ~~Adds `tg-force-scroll-reset` as an optional attribute on remotes. `tg-force-scroll-reset='true'` will cause a scroll reset when the redirect navigation concludes.~~
- ~~Adds `tg-force-head-asset-insertion` as an optional attribute on remotes. `tg-force-head-asset-insertion='true'` will cause TG to perform head asset tracking just like a regular navigation, while still respecting partial replace rules that may have been set.~~
- Adds tests for scroll reset behaviour so that future contributors can better understand how it is intended to behave.
- Refactors with an emphasis on simplifying unnecessary forking and currying and iteratively rolling out proper sinon assertions (`assert @spy.called` => `assert.called @spy`)
- Leaves behaviour unchanged for other redirects 

## Reviewers: 
@edward 
@GoodForOneFare 
@qq99 
@Shopify/tnt 

**Edit**
Redid this from the ground up with a much simpler convention of simply counting redirects to another url to be non-partial replaces. No new syntax.

### DIFF
--- a/lib/assets/javascripts/turbograft/page.coffee
+++ b/lib/assets/javascripts/turbograft/page.coffee
@@ -16,18 +16,18 @@ Page.refresh = (options = {}, callback) ->
   else
     location.href
 
-  if options.response
-    options.partialReplace = true
-    options.onLoadFunction = callback
+  turboOptions = {
+    partialReplace: true,
+    exceptKeys: options.exceptKeys,
+    onlyKeys: options.onlyKeys,
+    updatePushState: options.updatePushState,
+    callback: callback
+  }
 
-    xhr = options.response
-    delete options.response
-    Turbolinks.loadPage null, xhr, options
+  if xhr = options.response
+    Turbolinks.loadPage null, xhr, turboOptions
   else
-    options.partialReplace = true
-    options.callback = callback if callback
-
-    Turbolinks.visit newUrl, options
+    Turbolinks.visit newUrl, turboOptions
 
 Page.open = ->
   window.open(arguments...)
--- a/lib/assets/javascripts/turbograft/response.coffee
+++ b/lib/assets/javascripts/turbograft/response.coffee
@@ -1,5 +1,6 @@
 class TurboGraft.Response
   constructor: (@xhr) ->
+    @redirectedTo = @xhr.getResponseHeader('X-XHR-Redirected-To')
 
   valid: -> @hasRenderableHttpStatus() && @hasValidContent()
 
@@ -17,7 +18,15 @@ class TurboGraft.Response
     else
       throw new Error("Error encountered for XHR Response: #{this}")
 
+  redirectedToNewUrl: () ->
+    Boolean(
+      @redirectedTo &&
+      @redirectedTo != TurboGraft.location()
+    )
+
   toString: () ->
     "URL: #{@xhr.responseURL}, " +
     "ReadyState: #{@xhr.readyState}, " +
     "Headers: #{@xhr.getAllResponseHeaders()}"
+
+TurboGraft.location = () -> location.href
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -71,13 +71,14 @@ class window.Turbolinks
 
     rememberReferer()
 
-    options.partialReplace ?= false
-    options.onlyKeys ?= []
-    options.onLoadFunction = ->
-      resetScrollPosition() unless options.onlyKeys.length
-      options.callback?()
+    fetchReplacement(url, options)
 
-    fetchReplacement url, options
+  isPartialReplace = (response, options) ->
+    Boolean(
+      options.partialReplace ||
+      options.onlyKeys?.length ||
+      options.exceptKeys?.length
+    ) && !response.redirectedToNewUrl()
 
   @fullPageNavigate: (url) ->
     if url?
@@ -136,24 +137,30 @@ class window.Turbolinks
 
   @loadPage: (url, xhr, options = {}) ->
     triggerEvent 'page:receive'
+    response = new TurboGraft.Response(xhr)
     options.updatePushState ?= true
-    if upstreamDocument = new TurboGraft.Response(xhr).document()
-      if options.partialReplace
-        reflectNewUrl url if options.updatePushState
-        updateBody(upstreamDocument, xhr, options)
-      else
-        turbohead = new TurboGraft.TurboHead(activeDocument, upstreamDocument)
-        if turbohead.hasAssetConflicts()
-          return Turbolinks.fullPageNavigate(url)
-        reflectNewUrl url if options.updatePushState
-        turbohead.waitForAssets().then((result) ->
-          updateBody(upstreamDocument, xhr, options) unless result?.isCanceled
-        )
-    else
+    options.partialReplace = isPartialReplace(response, options)
+
+    unless upstreamDocument = response.document()
       triggerEvent 'page:error', xhr
       Turbolinks.fullPageNavigate(url)
+      return
 
-  updateBody = (upstreamDocument, xhr, options) ->
+    if options.partialReplace
+      reflectNewUrl url if options.updatePushState
+      updateBody(upstreamDocument, response, options)
+      return
+
+    turbohead = new TurboGraft.TurboHead(activeDocument, upstreamDocument)
+    if turbohead.hasAssetConflicts()
+      return Turbolinks.fullPageNavigate(url)
+
+    reflectNewUrl url if options.updatePushState
+    turbohead.waitForAssets().then((result) ->
+      updateBody(upstreamDocument, response, options) unless result?.isCanceled
+    )
+
+  updateBody = (upstreamDocument, response, options) ->
     nodes = changePage(
       upstreamDocument.querySelector('title')?.textContent,
       removeNoscriptTags(upstreamDocument.querySelector('body')),
@@ -161,16 +168,17 @@ class window.Turbolinks
       'runScripts',
       options
     )
-    reflectRedirectedUrl(xhr) if options.updatePushState
-    options.onLoadFunction?()
+
+    reflectRedirectedUrl(response) if options.updatePushState
+    Turbolinks.resetScrollPosition() unless options.partialReplace
+
+    options.callback?()
     triggerEvent 'page:load', nodes
 
   changePage = (title, body, csrfToken, runScripts, options = {}) ->
     activeDocument.title = title if title
-    options.onlyKeys ?= []
-    options.exceptKeys ?= []
 
-    if options.onlyKeys.length
+    if options.onlyKeys?.length
       nodesToRefresh = [].concat(getNodesWithRefreshAlways(), getNodesMatchingRefreshKeys(options.onlyKeys))
       nodes = refreshNodes(nodesToRefresh, body)
       setAutofocusElement() if anyAutofocusElement(nodes)
@@ -178,7 +186,7 @@ class window.Turbolinks
     else
       refreshNodes(getNodesWithRefreshAlways(), body)
       persistStaticElements(body)
-      if options.exceptKeys.length
+      if options.exceptKeys?.length
         refreshAllExceptWithKeys(options.exceptKeys, body)
       else
         deleteRefreshNeverNodes(body)
@@ -307,12 +315,11 @@ class window.Turbolinks
       Turbolinks.pushState { turbolinks: true, url: url.absolute }, '', url.absolute
     return
 
-  reflectRedirectedUrl = (xhr) ->
-    if location = xhr.getResponseHeader 'X-XHR-Redirected-To'
-      location = new ComponentUrl location
-      preservedHash = if location.hasNoHash() then activeDocument.location.hash else ''
-      Turbolinks.replaceState currentState, '', location.href + preservedHash
-
+  reflectRedirectedUrl = (response) ->
+    if url = response.redirectedTo
+      url = new ComponentUrl(url)
+      preservedHash = if url.hasNoHash() then activeDocument.location.hash else ''
+      Turbolinks.replaceState(currentState, '', url.href + preservedHash)
     return
 
   rememberReferer = ->
@@ -327,7 +334,7 @@ class window.Turbolinks
   recallScrollPosition = (page) ->
     window.scrollTo page.positionX, page.positionY
 
-  resetScrollPosition = ->
+  @resetScrollPosition: ->
     if activeDocument.location.hash
       activeDocument.location.href = activeDocument.location.href
     else
--- a/test/javascripts/response_test.coffee
+++ b/test/javascripts/response_test.coffee
@@ -1,5 +1,7 @@
 describe 'TurboGraft.Response', ->
   sandbox = null
+  iframe = null
+  baseHTML = '<html><head></head><body></body></html>'
 
   responseForFixture = (fixture, callback) ->
     xhr = new XMLHttpRequest
@@ -10,7 +12,15 @@ describe 'TurboGraft.Response', ->
     xhr.onerror = ->
       callback(new TurboGraft.Response(xhr))
 
+  setupIframe = ->
+    iframe = document.createElement('iframe')
+    document.body.appendChild(iframe)
+    iframe.contentDocument.write(baseHTML)
+    iframe.contentDocument
+
   beforeEach ->
+    testDocument = setupIframe() unless iframe
+    Turbolinks.document(testDocument)
     sandbox = sinon.sandbox.create()
     sandbox.useFakeServer()
     Object.keys(ROUTES).forEach (url) ->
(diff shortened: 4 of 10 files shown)


TheMallen:
## 🎩  Instructions (Shopify)
- go to `setup_shipping_card.rb`
- make it always show (`qualifies?` should just return true)
- go click the card's button and note the full page reset
- change line 24:
 ```
      def buttons
          [{
            text: 'Edit shipping settings',
            href: admin_dashboard_review_shipping_path,
            options: {
              data: {
                tg_remote: :post,
              }
            }
          }]
```
- try the card again, note that you no longer have a full page refresh, however you should have a js error and your scroll will not be reset properly
- go play around in orders and products and note that scroll position remains unchanged when you save a product or order
- fetch this branch of TG
- point your Shopify/Shopify project at your local copy
- change line 24:
```
      def buttons
          [{
            text: 'Edit shipping settings',
            href: admin_dashboard_review_shipping_path,
            options: {
              data: {
                tg_remote: :post,
                tg_force_scroll_reset: 'true',
                tg_force_head_asset_insertion: 'true'
              }
            }
          }]
```
- observe the scroll reset in action and the lack of a js error
- go play around with editing products and orders and notice that their behaviours are unchanged

File: lib/assets/javascripts/turbograft/initializers.coffee [hunk 1]
```diff
@@ -16,6 +16,8 @@ setupRemoteFromTarget = (target, httpRequestType, form = null) ->
   options =
     httpRequestType: httpRequestType
     httpUrl: httpUrl
+    forceScrollReset: TurboGraft.getTGAttribute(target, 'force-scroll-reset')
+    forceHeadAssetInsertion: TurboGraft.getTGAttribute(target, 'force-head-asset-insertion')
```
GoodForOneFare: It feels like these are always going to go together.  Is there situations in which `forceScrollReset != forceHeadAssetInertion`?

TheMallen: That's fair.

I might have been overdesigning the api, I just saw them as seperate things.

Maybe we can group them into a more semantic name as well?

GoodForOneFare: Yah, something higher level.  `tg_remote_navigation: "post"`, or something.

Reading the [tg-remote docs](https://github.com/Shopify/turbograft#data-tg-remote), I _think_ there's an implicit "this is for updating a form" message in there.  I think it's fair to make a distinction between in-place updates and submit-and-navigate actions.

qq99:
Another redirect scenario to think of is `next_redirect_to` aka the `X-Next-Redirect`
I'll have to think again on why we're using that methodology at all

TheMallen:
rebasing broke all the things :(

TheMallen:
## New 🎩  instructions
- go to `setup_shipping_card.rb`
- make it always show (`qualifies?` should just return true)
- go click the card's button and note the full page reset
- change line 24:
 ```
      def buttons
          [{
            text: 'Edit shipping settings',
            href: admin_dashboard_review_shipping_path,
            options: {
              data: {
                tg_remote: :post,
              }
            }
          }]
```
- try the card again, note that you no longer have a full page refresh.
  - scroll should be reset
  - no js errors should occur
- go play around in orders and products and note that scroll position remains unchanged when you save a product or order

**Comment by GoodForOneFare** [comment]
🎩 instructions check out.  Page updates without full refresh, and scrolling resets ❤️

I also poked around with creating/updating/duplicating orders, creating/updating/duplicating/deleting products, and creating transfers.  Everything seems to work fine.

File: lib/assets/javascripts/turbograft/page.coffee [hunk 2]
```diff
@@ -15,18 +15,15 @@ Page.refresh = (options = {}, callback) ->
     location.pathname + paramString
   else
     location.href
+  delete options.url
```
WillsonSmith: I'm always wary of `delete`, it doesn't remove references.

lib/assets/javascripts/turbograft/page.coffee
```diff
@@ -15,18 +15,15 @@ Page.refresh = (options = {}, callback) ->
...
 
+  options.partialReplace = true
+  options.callback = callback if callback
   if options.response
-    options.partialReplace = true
-    options.onLoadFunction = callback
-
     xhr = options.response
     delete options.response
```
WillsonSmith: I just noticed this is here, does this remove references properly? Pretty sure `delete` will leave any references to the object deleted. Is setting `xhr` to `options.response` and then deleting it leaking?

lib/assets/javascripts/turbograft/response.coffee
```diff
@@ -17,7 +17,17 @@ class TurboGraft.Response
     else
       throw new Error("Error encountered for XHR Response: #{this}")
 
+  redirectedTo: () ->
+    @_redirectedTo = @xhr.getResponseHeader('X-XHR-Redirected-To') unless @_redirectedTo
+    @_redirectedTo
```
WillsonSmith: I think this will be returned by the first line.

lib/assets/javascripts/turbograft/page.coffee
WillsonSmith: maybe 
```coffee
options.url = null
delete options.url
```

lib/assets/javascripts/turbograft/response.coffee
```diff
@@ -17,7 +17,17 @@ class TurboGraft.Response
...
+
+  redirectedToNewUrl: () ->
+    @redirectedTo() && (@redirectedTo() != TurboGraft.Response.currentLocation())
```
WillsonSmith: I don't really like relying on `undefined` for a conditional. That's what `@redirectedTo()` will return if not the response header, right?

lib/assets/javascripts/turbograft/response.coffee
TheMallen: True

TheMallen: Hmm

Ok, so this was probably just over optimizing.

 I was intending to memoize the call to `getResponseHeader` onto the response object. That's why it's written this way (we can't just return the result of the assignment if we are only doing the assignment when it's not set...)

I think I'll just get rid of the memoization.

lib/assets/javascripts/turbograft/page.coffee
```diff
@@ -16,18 +16,18 @@ Page.refresh = (options = {}, callback) ->
...
+  turboOptions = {
+    partialReplace: true,
+    exceptKeys: options.exceptKeys,
+    onlyKeys: options.onlyKeys,
+    updatePushState: options.updatePushState,
+    callback: callback
+  }
 
-    xhr = options.response
-    delete options.response
```
qq99: I wonder why we were doing this?

lib/assets/javascripts/turbograft/response.coffee
```diff
@@ -17,7 +17,19 @@ class TurboGraft.Response
     else
       throw new Error("Error encountered for XHR Response: #{this}")
 
+  redirectedTo: () ->
+    @xhr.getResponseHeader('X-XHR-Redirected-To')
+
+  redirectedToNewUrl: () ->
+    Boolean(
+      @redirectedTo() &&
```
qq99: memoize `redirectedTo()` maybe

lib/assets/javascripts/turbograft/response.coffee
```diff
@@ -17,7 +17,19 @@ class TurboGraft.Response
     else
       throw new Error("Error encountered for XHR Response: #{this}")
 
+  redirectedTo: () ->
+    @xhr.getResponseHeader('X-XHR-Redirected-To')
+
+  redirectedToNewUrl: () ->
+    Boolean(
```
qq99: is this necessary? I thought `foo && bar` was always a boolean result

qq99: screen shot 2016-10-11 at 12 03 15 pm

😢 nvm lol

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -136,39 +136,48 @@ class window.Turbolinks
 
   @loadPage: (url, xhr, options = {}) ->
     triggerEvent 'page:receive'
+    response = new TurboGraft.Response(xhr)
+    options.onlyKeys ?= []
+    options.exceptKeys ?= []
```
qq99: imo this feels a bit weird to set here

Shouldn't it be pretty unlikely now that this is not defined?  What if `isPartialReplace` did `options.exceptKeys?.length`?

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -136,39 +136,48 @@ class window.Turbolinks
...
+
+    options.callback?()
     triggerEvent 'page:load', nodes
 
   changePage = (title, body, csrfToken, runScripts, options = {}) ->
     activeDocument.title = title if title
-    options.onlyKeys ?= []
-    options.exceptKeys ?= []
 
     if options.onlyKeys.length
```
qq99: similarly, I think it might have been a historical mistake to have `options.onlyKeys ?= []` above, if only for this check to pass without exception -- `options.onlyKeys?.length`

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -307,12 +316,11 @@ class window.Turbolinks
...
-  reflectRedirectedUrl = (xhr) ->
-    if location = xhr.getResponseHeader 'X-XHR-Redirected-To'
-      location = new ComponentUrl location
-      preservedHash = if location.hasNoHash() then activeDocument.location.hash else ''
-      Turbolinks.replaceState currentState, '', location.href + preservedHash
-
+  reflectRedirectedUrl = (response) ->
+    if url = response.redirectedTo()
+      url = new ComponentUrl(url)
+      preservedHash = if url.hasNoHash() then document.location.hash else ''
```
qq99: not `activeDocument`?

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -136,39 +136,48 @@ class window.Turbolinks
...
+      reflectNewUrl url if options.updatePushState
+      updateBody(upstreamDocument, response, options)
+      return
 
-  updateBody = (upstreamDocument, xhr, options) ->
+    turbohead = new TurboGraft.TurboHead(activeDocument, upstreamDocument)
+    if turbohead.hasAssetConflicts()
+      return Turbolinks.fullPageNavigate(url)
+
+    reflectNewUrl url if options.updatePushState
```
qq99: is it cool that `reflectNewUrl` and `reflectRedirectedUrl` can both fire? this here, and then `reflectRedirectedUrl` in the `updateBody` call?

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -71,14 +71,14 @@ class window.Turbolinks
...
 
-    options.partialReplace ?= false
-    options.onlyKeys ?= []
-    options.onLoadFunction = ->
-      resetScrollPosition() unless options.onlyKeys.length
-      options.callback?()
-
     fetchReplacement url, options
 
+  isPartialReplace = (options, response) ->
```
qq99: I wonder if `isPartialReplace` might better belong inside of `Response`, or not. It seems like `turbolinks.coffee` only instantiates it when `options` is also in scope, so `options` could be passed in too.  Is it a Response responsibility?

lib/assets/javascripts/turbograft/response.coffee
TheMallen: I was doing that but it was a little overcomplicated and weird.

lib/assets/javascripts/turbograft/turbolinks.coffee
TheMallen: Good call.

TheMallen: Probably a rebasing error on my part.

lib/assets/javascripts/turbograft/turbolinks.coffee
TheMallen: I think it's a `visit` responsibility. We just don't have a `visit` object yet ;)

lib/assets/javascripts/turbograft/turbolinks.coffee
TheMallen: 👍 I like this plan.

lib/assets/javascripts/turbograft/response.coffee
lemonmade: I don't see why it was weird, couldn't you just have `redirectedTo = @xhr.getResponseHeader('X-XHR-Redirected-To')` outside of this `Boolean` call and use it in place of the method call?

lib/assets/javascripts/turbograft/response.coffee
```diff
@@ -17,7 +17,19 @@ class TurboGraft.Response
...
+      @redirectedTo() &&
+      @redirectedTo() != TurboGraft.Response.currentLocation()
```
lemonmade: Seems unecessary to have a method to get this when you could just access the current location (unless this is for a test...)

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -71,14 +71,14 @@ class window.Turbolinks
...
+    Boolean(options.partialReplace ||
+    options.onlyKeys?.length ||
+    options.exceptKeys?.length) &&
```
lemonmade: This is hard to read with the lack of indentation/ arguments not being lined up.

lib/assets/javascripts/turbograft/turbolinks.coffee
lemonmade: I kind of agree that in the lack of a larger refactor, it feels like this should be in the response (especially since half the logic is already in the response).

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -136,49 +136,56 @@ class window.Turbolinks
...
-      else
-        turbohead = new TurboGraft.TurboHead(activeDocument, upstreamDocument)
-        if turbohead.hasAssetConflicts()
-          return Turbolinks.fullPageNavigate(url)
-        reflectNewUrl url if options.updatePushState
-        turbohead.waitForAssets().then((result) ->
-          updateBody(upstreamDocument, xhr, options) unless result?.isCanceled
-        )
-    else
+    options.partialReplace = isPartialReplace(options, response)
```
lemonmade: It feels gross to assign to an object, using that object as an input into the function (also, options should always come last in a function IMO)

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -307,12 +314,11 @@ class window.Turbolinks
...
     return
 
-  reflectRedirectedUrl = (xhr) ->
-    if location = xhr.getResponseHeader 'X-XHR-Redirected-To'
-      location = new ComponentUrl location
-      preservedHash = if location.hasNoHash() then activeDocument.location.hash else ''
-      Turbolinks.replaceState currentState, '', location.href + preservedHash
-
+  reflectRedirectedUrl = (response) ->
+    if url = response.redirectedTo()
```
lemonmade: Aha, here's another use of that method. Maybe it should just be a property, since it can be determined in `Response`'s constructor?

test/javascripts/response_test.coffee
```diff
@@ -55,3 +65,28 @@ describe 'TurboGraft.Response', ->
...
+
+  describe 'redirectedTo', ->
+    it 'returns the value of the X-XHR-Redirected-To header', (done) ->
+      responseForFixture 'xhrRedirectedToHeader', (response) ->
+        assert.equal(response.redirectedTo(), ROUTES['xhrRedirectedToHeader'][1]['X-XHR-Redirected-To'])
+        done()
+
+  describe 'redirectedToNewUrl', ->
+    beforeEach ->
+      sandbox.stub(TurboGraft.Response, 'currentLocation', -> 'test-location')
```
lemonmade: I guess you can't stub the native window location, but this doesn't feel like the right place to have a method for accessing this. I would say either 1. Force the consuming method to accept a location, so that you can just pass it in instead of reading the global, or 2) Add another object that has a method for returning the current location that is more appropriate than the response.

TheMallen:
Talked to chris in person, he ok'd the last commit :)

TheMallen:
Aw yiss, squash time

image

TheMallen:
🎉

PR #163, opened 2016-09-29
turbolinks accepts a document, run tests inside an iframe

### LINKED ISSUE #138: Turbolinks should accept a document parameter

TheMallen:
## We now have a browser runner again!
image
(closes #138)

This PR makes two primary changes.
- `Turbolinks.document` allows you to set a document for turbolinks to operate on.
- Tests are now run leveraging this, allowing for use of the browser test runner by sandboxing TG's DOM mutations to an iframe.

## How
We had to move the way we defined ASSET_FIXTURES due to some weirdness in teaspoon config, and change all uses of `document` to reference a variable in module scope to facilitate this change. We also had to make some minor changes in how we initialize tests.

## Why
Development velocity is better when we can actually use the browser test runner. Everything else is gravy.

## Follow up:
We should change `remote_test` to do this as well, since it's currently suffering from the same problem `turbolinks_test` had.

## Reviewers
@qq99 
@GoodForOneFare 
@edward 
CC @Shopify/tnt


qq99:
Nice :D

test/javascripts/turbolinks_test.coffee
```diff
@@ -91,23 +101,32 @@ describe 'Turbolinks', ->
...
     $("#turbo-area").remove()
 
   it 'is defined', ->
     assert(Turbolinks)
 
+  describe '#document', ->
+    afterEach ->
+      Turbolinks.document(testDocument)
+
+    it 'assigning a document', ->
```
GoodForOneFare: Rename to state some kind of behaviour, or cause and effect.  Or just delete.

GoodForOneFare:
I can run browser tests!  😃 Excellent stuff 👍

test/javascripts/turbolinks_test.coffee
TheMallen: I'll delete it :P

TheMallen:
@qq99 You think this is good to 🚢 ?

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -42,7 +43,7 @@ requestMethodIsSafe =
 browserSupportsTurbolinks = browserSupportsPushState and requestMethodIsSafe
 
 browserSupportsCustomEvents =
-  document.addEventListener and document.createEvent
+  activeDocument.addEventListener and activeDocument.createEvent
```
qq99: this could probably just be `document` right?

PR #164, opened 2016-10-04
Version bump to 0.4.3

GoodForOneFare:
This is mainly to get the fix for meaningless null exceptions out of Bugsnag.  Everything else is very nice gravy (thanks for acting like an owner, @TheMallen).

Please review: @TheMallen, @utkarshsaxenashopify


TheMallen:
We should probably write some patch notes. :P

TheMallen:
(lgtm tho)

utkarshsaxenashopify:
👍

PR #167, opened 2016-10-20
client side redirect detection

TheMallen:
## Summary
- adds a new check that determines redirect status on the client side using `xhr.responseURL` vs our known`intendedURL`
- memoizes `url` onto the `Response` object 
- simplifies browser state updates using `Response.url`
## Why tho

Basically, in some cases we were causing the server to do extra work. It turns out this becomes more apparent when service workers get involved, because you need to know what URL you're actually going to and you can't know it will be a redirect first and... ok well it's hard to explain why but let me tell you it was breaking something fierce.

Imagine if you will, you go to do a thing, and you get redirected. Lucky you, this also happens to be right when we push a new bundle. ATM the following happens:
- TG makes an xhr to the url you wanted, and stores this url
- The server does things and sends a redirect
- Browser follows this automagically
- TG receives the response from the redirect
- TG checks the head assets and decides to reload the page **BUT, it uses the original url you wanted**
- The Browser navigates to the url you wanted
- The server does things (again) and sends a redirect (again, maybe, hopefully I guess)
- The Browser follows this automagically

This usually hasn't been an issue for us because most of our redirects are through `Remote` which... doesn't give a url to `Page`. This is probably why we used to have so many exceptions out of `Turbolinks` related to trying to `fullPageNavigate` to a null url. Anyways, it breaks my service worker implementation (causes an infinite redirection loop actually) as well as just being wasteful and possibly triggering server work multiple times for no reason.
## What why do we need `X-XHR-Redirected-To` then?

The short answer is we don't. The long answer is that `Remote` is hiding the url it's trying to go to from our friends `Page` and `Turbolinks`. It could be a large-ish refactor involving a test murdering spree to make that not the case, so I'll leave that for the future for now. 
## Wait why did you remove the use of `reflectRedirectedUrl`

As far as I could tell the flow was basically
- always push the new state onto the history stack (can happen in 2 places)
- if we have a redirect header replace the top entry with the redirect header's url (later)

The result of this is the same as
- determine the canonical url to push
- push the url
## Tophatting

I tophatted a lot with these changes. But it should be tophatted more for sure. I'd recommend going to lots of forms, logging out and heading back to pages of the admin, going to theme editor, and just generally making a muck of things and causing redirects allover the place.
## Reviewers

@lemonmade 
@qq99
@qq99

CC @Shopify/tnt

### DIFF
--- a/lib/assets/javascripts/turbograft/response.coffee
+++ b/lib/assets/javascripts/turbograft/response.coffee
@@ -1,6 +1,11 @@
 class TurboGraft.Response
-  constructor: (@xhr) ->
-    @redirectedTo = @xhr.getResponseHeader('X-XHR-Redirected-To')
+  constructor: (@xhr, intendedURL) ->
+    if intendedURL && intendedURL != @xhr.responseURL
+      @redirectedTo = @xhr.responseURL
+    else
+      @redirectedTo = @xhr.getResponseHeader('X-XHR-Redirected-To')
+
+    @url = @redirectedTo || intendedURL
 
   valid: -> @hasRenderableHttpStatus() && @hasValidContent()
 
--- a/lib/assets/javascripts/turbograft/response.coffee
+++ b/lib/assets/javascripts/turbograft/response.coffee
@@ -5,7 +5,7 @@ class TurboGraft.Response
     else
       @redirectedTo = @xhr.getResponseHeader('X-XHR-Redirected-To')
 
-    @url = @redirectedTo || intendedURL
+    @finalURL = @redirectedTo || intendedURL
 
   valid: -> @hasRenderableHttpStatus() && @hasValidContent()
 
--- a/lib/assets/javascripts/turbograft/response.coffee
+++ b/lib/assets/javascripts/turbograft/response.coffee
@@ -1,6 +1,6 @@
 class TurboGraft.Response
   constructor: (@xhr, intendedURL) ->
-    if intendedURL && intendedURL != @xhr.responseURL
+    if intendedURL && intendedURL.withoutHash() != @xhr.responseURL
       @redirectedTo = @xhr.responseURL
     else
       @redirectedTo = @xhr.getResponseHeader('X-XHR-Redirected-To')
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -151,7 +151,6 @@ class window.Turbolinks
       return
 
     if options.partialReplace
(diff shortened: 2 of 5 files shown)


TheMallen:
Looks like travis is trying to use Selenium for some reason and exploding. Strange.

TheMallen:
@GoodForOneFare 
@qq99

Pls 2 take a look at this :)

GoodForOneFare:
Makes sense 👍 This gave us so many brain itches!  Thanks for taking ownership and figuring out a simpler flow.

GoodForOneFare:
I'll find some time to 🎩 by the end of the week.  Got any tips for what/how to cover this, @TheMallen?

TheMallen:
Stuff where there are redirects, so cards, form actions, going to theme editor are the main ones.

TheMallen:
@qq99 Should be fixed now, could you confirm? :)

lib/assets/javascripts/turbograft/response.coffee
```diff
@@ -1,6 +1,11 @@
 class TurboGraft.Response
-  constructor: (@xhr) ->
-    @redirectedTo = @xhr.getResponseHeader('X-XHR-Redirected-To')
+  constructor: (@xhr, intendedURL) ->
+    if intendedURL && intendedURL.withoutHash() != @xhr.responseURL
```
qq99: `@xhr.responseURL` can this ever contain a hash?

TheMallen: https://www.fxsitecompat.com/en-CA/docs/2014/xhr-responseurl-no-longer-includes-fragment/

TheMallen:
AW yiss

GoodForOneFare:
Did an Admin odyssey in Chrome, Firefox, Safari, and IE11.  Tested: theme editor navigation; redirect to logout (in FF + IE11), creating/updating orders + products.  Navigating around payments, domains, settings in various browsers, and things looked cool.

A couple of times, navigating from a Home card using tg-remote caused the shipping page to display without downloading `admin/index`, but this only happened twice in Chrome + once in Safari.  I'm not sure what the reproduction steps are, even 😖

GoodForOneFare:
TL;DR: 👍 shipit!

I've only managed to reproduce this issue twice since my original comment.  First time was after switching back from an esified branch, second time was after adding TurboGraft debug statements.  That points at asset changes being a factor, but me and @TheMallen poked around with that and the system diligently performs a full page refresh on asset change (even for home card->shipping navigation).  I think this was just my server getting into a screwy state.

PR #168, opened 2016-10-20
refactor remote tests to use iframe

TheMallen:
This test makes remote test browser runner friendly by sandboxing TG into an iframe just like in the other tests. :)

@Shopify/tnt 
@GoodForOneFare 
@lemonmade

### DIFF
--- a/test/javascripts/remote_test.coffee
+++ b/test/javascripts/remote_test.coffee
@@ -1,7 +1,18 @@
+baseHTML = '<html><head></head><body></body></html>'
+iframe = null
+
+setupIframe = ->
+  iframe = document.createElement('iframe')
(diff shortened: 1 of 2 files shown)


test/javascripts/remote_test.coffee
```diff
@@ -1,7 +1,18 @@
...
+  iframe = document.createElement('iframe')
+  document.body.appendChild(iframe)
+  iframe.contentDocument.write(baseHTML)
+  iframe.contentDocument
+
 describe 'Remote', ->
   @initiating_target = null
 
   beforeEach ->
+    testDocument = setupIframe() unless iframe
```
GoodForOneFare: Does this mean that the previous test's url/content will leak into the current test?

PR #170, opened 2016-10-21
Fix broken travis scripts, plug up ensuing test leaks.

### LINKED ISSUE #169: Our test runner isn't running
Our CI fails on trying to start selenium atm.

Not sure why it's trying to start selenium. Maybe something was changed on the travis side? I think the last time we touched the testing config in the repo was months ago. Actually how could a PR to make that change even get through CI if it makes CI fail?

Yeah I'd bet it was something on the travis config side.

@Shopify/tnt 
@qq99 
@GoodForOneFare

TheMallen:
**Fixes #169 (hopefully?)**
~~We just export an environment variable. TBH I'm not sure how this was ever working if travis never had selenium.~~

This was funny. 
- Travis suddenly lacked selenium drivers for FF
- Tried to switch to phantom
  - Worked great for js tests
  - Didn't work for ruby tests
- Switched ruby tests back to selenium and worked on getting it working
  - Had to install geckodriver in the before_script hook
  - Had to request a modern firefox build
- Ran into leaky tests (possibly exposed by newer FF build or something?) where test order randomization would cause fails sometimes due to session persistence and our server side TG shenanigans.
  - Added `session_reset` logic to fix them
- Had a hilarious number of fails (and thus test restarts) caused by our silly test that compares two random numbers to eachother. I'm not really sure how to fix that test without a larger test restructure so I'm going to leave it for now. (*sigh*). In theory it should be like a million in one chance but there must be something funky happening in the RNG to make it more likely. (shared seeds?)

@Shopify/tnt 
@qq99 
@lemonmade
@GoodForOneFare

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,25 @@
+addons:
+  firefox: "49.0.2"
+
 language: ruby
+sudo: false
+
+cache:
+  directories:
+    - "travis_geckodriver"
+
 rvm:
   - '2.3.1'
+
 before_script:
+  - if [ ! -d $PWD/travis_geckodriver ];  then mkdir $PWD/travis_geckodriver; fi
+  - if [ ! -f $PWD/travis_geckodriver/geckodriver ]; then wget https://github.com/mozilla/geckodriver/releases/download/v0.11.1/geckodriver-v0.11.1-linux64.tar.gz; fi
+  - if [ ! -f $PWD/travis_geckodriver/geckodriver ]; then tar -xvf geckodriver-v0.11.1-linux64.tar.gz -C $PWD/travis_geckodriver; fi
+  - "export PATH=$PWD/travis_geckodriver:$PATH"
+  - "geckodriver --version"
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
   - sleep 3
   - bundle install
+
 script: bundle exec rake ci
--- a/test/example/app/assets/javascripts/application.js
+++ b/test/example/app/assets/javascripts/application.js
@@ -12,5 +12,6 @@
 //
 //= require jquery
 //= require jquery_ujs
+//= require_tree ./vendor
 //= require turbograft
 //= require_tree .
--- a/turbograft.gemspec
+++ b/turbograft.gemspec
@@ -6,10 +6,10 @@ require 'turbograft/version'
 Gem::Specification.new do |spec|
   spec.name          = "turbograft"
   spec.version       = TurboGraft::VERSION
-  spec.authors       = ["Kristian Plettenberg-Dussault", "Justin Li", "Nicholas Simmons", "Tyler Mercier", "Anthony Cameron", "Patrick Donovan"]
-  spec.email         = ["tylermercier@gmail.com"]
+  spec.authors       = ["Kristian Plettenberg-Dussault", "Justin Li", "Nicholas Simmons", "Tyler Mercier", "Anthony Cameron", "Patrick Donovan", "Mathew Allen", "Gord Pearson"]
+  spec.email         = ["tylermercier@gmail.com", "mathew.allen@shopify.com"]
   spec.summary       = "turbolinks with partial page replacement"
-  spec.description   = "It's like turbolinks, but with partial page replacement and tests"
+  spec.description   = "Turbograft is a hard fork of Turbolinks, allowing you to perform partial page refreshes and offering ajax form utilities."
   spec.homepage      = "https://github.com/Shopify/turbograft"
   spec.license       = "MIT"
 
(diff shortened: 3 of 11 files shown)


GoodForOneFare:
Weird.  LGTM 👍 Wish I knew why this was happening.

TheMallen:
Stuff I did:
- Manually followed the test codepaths in my browser for each failing test (behaved correctly in real browsers)
- Got selenium working on my machine finally (all tests passed)

image

**It seems to me that poltergeist is the cause of the breakages**. I'll try to get selenium working on travis.

TheMallen:
image
 I am the winningest

TheMallen:
@GoodForOneFare
@lemonmade 
@qq99

This fixes broken tests. I think it also removes basically all test flakiness except for the stupid random number one (which incidentally I ran into like 10 times while doing this).

If I get some 👍  I can squash and merge.

TheMallen:
Eh I'm just gonna squash it now.

.travis.yml
```diff
@@ -1,9 +1,34 @@
...
+
 rvm:
   - '2.3.1'
+
 before_script:
+  - "phantomjs --version"
+  - "export PATH=$PWD/travis_phantomjs/phantomjs-2.1.1-linux-x86_64/bin:$PATH"
+  - "phantomjs --version"
+  - "if [ $(phantomjs --version) != '2.1.1' ]; then rm -rf $PWD/travis_phantomjs; mkdir -p $PWD/travis_phantomjs; fi"
+  - "if [ $(phantomjs --version) != '2.1.1' ]; then wget https://assets.membergetmember.co/software/phantomjs-2.1.1-linux-x86_64.tar.bz2 -O $PWD/travis_phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2; fi"
```
GoodForOneFare: Did the previous line just delete phantomjs?  If `phantomjs` is no longer a valid command, is this ever going to execute properly?

turbograft.gemspec
```diff
@@ -6,7 +6,7 @@ require 'turbograft/version'
 Gem::Specification.new do |spec|
   spec.name          = "turbograft"
   spec.version       = TurboGraft::VERSION
-  spec.authors       = ["Kristian Plettenberg-Dussault", "Justin Li", "Nicholas Simmons", "Tyler Mercier", "Anthony Cameron", "Patrick Donovan"]
+  spec.authors       = ["Kristian Plettenberg-Dussault", "Justin Li", "Nicholas Simmons", "Tyler Mercier", "Anthony Cameron", "Patrick Donovan", "Mathew Allen"]
```
GoodForOneFare: 🎉

TheMallen: Haha I just realized this is still here...

I had modified the gemspec when I thought we would use poltergeist  so I threw in this change... Now it just looks random and braggy. :P

TheMallen: You should be there too ^.^

.travis.yml
TheMallen: `phantomjs` will always be a valid command because the older version is installed on travis by default. 

This might want to be rethought since it is pretty weird I guess.

GoodForOneFare:
LGTM 👍

turbograft.gemspec
```diff
@@ -6,7 +6,7 @@ require 'turbograft/version'
...
   spec.email         = ["tylermercier@gmail.com"]
```
qq99: we should prob change this too eh

turbograft.gemspec
```diff
@@ -6,7 +6,7 @@ require 'turbograft/version'
...
   spec.summary       = "turbolinks with partial page replacement"
   spec.description   = "It's like turbolinks, but with partial page replacement and tests"
```
qq99: doh, can you also remove "and tests" here?

test/example/app/assets/javascripts/application.js
```diff
@@ -12,5 +12,6 @@
 //
 //= require jquery
 //= require jquery_ujs
+//= require_tree ./vendor
```
qq99: ? what do you need this and promise.js for?

turbograft.gemspec
TheMallen: Tru

turbograft.gemspec
TheMallen: OMG YOU FOUND IT

I noticed our overly smarmy description on rubygems and it made me sad but I couldn't figure out where it was coming from. Never occurred to me it would be in the gemspec. :{

.travis.yml
```diff
@@ -1,9 +1,25 @@
+addons:
+  firefox: "latest"
```
TheMallen: Good catch.

TheMallen:
image

TheMallen:
🚢 ing

TheMallen:
Squashing broke it?!

Issue #159, opened 2016-09-23
Still has tests that are flaky


TheMallen:
I am sad.

`test/browser/legacy_partial_page_refresh_test.rb` seems to be flaky, I merged a PR today which was green, master became red, I reran travis tests on master, master became green.

TheMallen:
Found the flaky tests:
https://github.com/Shopify/turbograft/blob/963bd8b4686ac8a1c18d0b2b256db65b9dfc908a/test/browser/legacy_partial_page_refresh_test.rb

Comparing two random numbers is a bad strategy, no garauntee they won't randomly be the same. Also, `assert`ing the results of a `Capybara` find is pointless because it will raise an exception if it fails to find.

We'll need to fix all the tests that look like this if we want to prevent pointless reruns.

PR #173, opened 2016-11-10
Make ruby tests not randomly fail

TheMallen:
**Fixes** #159 

## Problem

The TG sample app has two `rand(1000)` calls inside of our templates to create random numbers on every request to the server. The partial replacement tests were using these to check that partial page replacement was working properly. 

Naturally, this causes the occasional test fail because when you do the partial replace your new random is sometimes equal to your old random. 

You might naively think this should only very rarely occur, since the probability of two true random numbers between 1 and 1000 should be like, one in a million. Putting aside my opinion that one in a million tests failing for silly reasons is too many, the fact that we had two files of tests where every testcase essentially asserted `rand(1000) != rand(1000)` (sometimes several times) meant that we encountered failed tests often enough to already lose me a few hours of rerunning tests over the last couple months.

## Fix
We use session to track two counters and iterate them on each navigation. Since we clear session between each testcase, and since they only ever increase the counters, we can know that the only reason a number would be equal when we don't expect it to be would be because our functionality is actually broken.

## Reviewers
@qq99 
@GoodForOneFare 
@Shopify/tnt

### DIFF
--- a/test/example/app/controllers/pages_controller.rb
+++ b/test/example/app/controllers/pages_controller.rb
@@ -1,4 +1,5 @@
 class PagesController < ApplicationController
+  before_action :setup_counters
   skip_before_action :verify_authenticity_token
 
   def index
@@ -78,5 +79,17 @@ class PagesController < ApplicationController
 
   def method_agnostic
     render json: {method: request.env["REQUEST_METHOD"]}, status: 200
+  end
+
+  private
+  def setup_counters
+    [:counter_a, :counter_b].each do |counter|
+      setup_or_iterate_counter!(counter)
+    end
+  end
+
+  def setup_or_iterate_counter!(counter)
+    session[counter] ||= 0
+    session[counter] += 1
   end
 end
--- a/test/example/app/controllers/pages_controller.rb
+++ b/test/example/app/controllers/pages_controller.rb
@@ -1,5 +1,4 @@
 class PagesController < ApplicationController
-  before_action :setup_counters
   skip_before_action :verify_authenticity_token
 
   def index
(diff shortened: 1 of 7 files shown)


TheMallen:
Here's what the sample app looks like now, btw:
image

test/example/app/controllers/pages_controller.rb
```diff
@@ -79,4 +80,16 @@ def new
...
+  def setup_counters
+    [:counter_a, :counter_b].each do |counter|
+      setup_or_iterate_counter!(counter)
+    end
+  end
+
+  def setup_or_iterate_counter!(counter)
+    session[counter] ||= 0
+    session[counter] += 1
+  end
```
qq99: could this be put into `ApplicationController`?

test/example/app/views/pages/show.html.erb
```diff
@@ -6,13 +6,15 @@
...
     <span>Section A:</span>
-    <span id="random-number-a"><%= rand(1000) %></span>
+    <p>Random Number: <span id="random-number-a"><%= rand(1000) %></span></p>
+    <p>Iterative Number: <span id="iterative-number-a"><%= session[:counter_a] %></span></p>
   </div>
-  <div id='section-b' data-tg-refresh='section-b' style="padding: 1em; margin: 1em; background: <%= random_light_rgb() %>; color: black;">
+  <div id='section-b' data-tg-refresh='section-b' style="padding: 1em; background: <%= random_light_rgb() %>; color: black;">
     <span>Section B:</span>
-    <span id="random-number-b"><%= rand(1000) %></span>
+    <p>Random Number: <span id="random-number-b"><%= rand(1000) %></span></p>
```
lemonmade: Are the random numbers still used in some tests?

TheMallen: No, but it's also a sample app you can navigate around so I thought maybe I should leave them to help illustrate how it works (assumedly their original purpose).

GoodForOneFare:
I'd go for `counter-a` / `counter-b`, but everything LGTM 👍 🚢

PR #174, opened 2016-12-12
Allow to configure Turbogradft in a way that doesn't compromise lazy loading

casperisfine:
Currently, if you customize `TurboGraft::Config.controllers`, the listed controllers will be eager loaded.

This is detrimental in big apps, because it will slow down the boot in development mode for no good reason.

This PR allows alternatively to explicitly `include TurboGraft` in the controllers you want it, and it will only take effect when the controller is referenced.

@qq99 @TheMallen @rafaelfranca for review please.

### DIFF
--- a/lib/turbograft.rb
+++ b/lib/turbograft.rb
@@ -14,16 +14,20 @@ module TurboGraft
     self.controllers = ["ActionController::Base"]
   end
 
+  def self.included(controller)
+    controller.class_eval do
+      include XHRHeaders, Cookies, XDomainBlocker, Redirection
+      before_action :set_xhr_redirected_to, :set_request_method_cookie
+      after_action :abort_xdomain_redirect
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


lib/turbograft.rb
```diff
@@ -14,16 +14,20 @@ class Config
     self.controllers = ["ActionController::Base"]
```
rafaelfranca: Maybe we should remove this from here too? We don't need to patch the framework controller since users can add the include in the `ApplicationController`

qq99:
cc @Shopify/tnt

lib/turbograft.rb
casperisfine: If it was just me I'd remove this way to hook into the app entirely. But It's a breaking change so I haven't proposed it.

But yeah I'd be fine with this.

casperisfine:
@Shopify/tnt thoughts?

PR #175, opened 2016-12-15
Have visit return a promise, add associated test

TheMallen:
## Problem
This PR adds promise support for running code after a replace has occurred. We currently offer other ways of doing this, such as passing a callback to `Page.replace` or adding an event listener for `page:load`, however these apis are more developer-hostile and in real life we've seen things like this:

```
# someform.coffee (line 200)
    operation = $.ajax(.....)

    operation.done (data, status, xhr) ->
        Page.refresh response: xhr

    @saveOperation = operation

# someform.coffee (line 300)
    @saveOperation.done (data) ->
        doSomeOperationsThatRequireTGToBeDoneDoingItsThing()
```

Developers keep the ajax object around so that later parts of the program can dynamically add listeners to them. **Unfortunately, this doesn't always work.** Since TG's `loadPage` operation is only synchronous when we don't have to update head assets, **you cannot garauntee that a later `done` callback on the ajax object will be executed after TG is done updating the page**. This can cause some crazy issues.

Of course, the dev could build their own promise or other object to manage `page:load` callbacks, but this is a lot of boilerplate.

## Solution
We are already sometimes returning a promise (more or less accidentally) from `loadPage`. Let's just always do so. This lets code simply switch to storing the promise result from `refresh` or `visit`, and involves minimal changes and no api breakage. In the future, we could even make `loadPage` work with `Response` objects returned by `fetch`.

```
# someform.coffee (line 300)
    @saveOperation.then (data) ->
        doSomeOperationsThatRequireTGToBeDoneDoingItsThing()
```

@lemonmade 
@GoodForOneFare 
@Shopify/byroot

### DIFF
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -70,7 +70,7 @@ class window.Turbolinks
   referer = null
 
   fetch = (url, options = {}) ->
-    return if pageChangePrevented(url)
+    return Promise.reject() if pageChangePrevented(url)
     url = new ComponentUrl(url)
 
     rememberReferer()
@@ -121,10 +121,11 @@ class window.Turbolinks
     for k,v of options.headers
       xhr.setRequestHeader k, v
 
-    new Promise (resolve) => 
+    new Promise (resolve, reject) =>
       xhr.onload = ->
         if xhr.status >= 500
           Turbolinks.fullPageNavigate(url)
+          reject()
         else
           resolve(Turbolinks.loadPage(url, xhr, options))
         xhr = null
@@ -134,6 +135,7 @@ class window.Turbolinks
         if xhr.statusText == "abort"
           xhr = null
           return
+
         Turbolinks.fullPageNavigate(url)
 
       xhr.send()
@@ -146,14 +148,14 @@ class window.Turbolinks
 
     unless upstreamDocument = response.document()
       triggerEvent 'page:error', xhr
-      return Turbolinks.fullPageNavigate(response.finalURL);
+      return Promise.reject(Turbolinks.fullPageNavigate(response.finalURL))
 
     if options.partialReplace
-      return Promise.resolve(updateBody(upstreamDocument, response, options));
+      return Promise.resolve(updateBody(upstreamDocument, response, options))
 
     turbohead = new TurboHead(activeDocument, upstreamDocument)
     if turbohead.hasAssetConflicts()
-      return Turbolinks.fullPageNavigate(response.finalURL)
+      return Promise.reject(Turbolinks.fullPageNavigate(response.finalURL))
 
     turbohead.waitForAssets().then((result) ->
       updateBody(upstreamDocument, response, options) unless result?.isCanceled
@@ -173,7 +175,7 @@ class window.Turbolinks
 
     options.callback?()
     triggerEvent('page:load', nodes)
-    
+
     Promise.resolve(nodes)
 
   changePage = (title, body, csrfToken, runScripts, options = {}) ->
@@ -201,7 +203,7 @@ class window.Turbolinks
       triggerEvent 'page:change'
       triggerEvent 'page:update'
 
-    return
+    return [].slice.call(body.children)
 
   getNodesMatchingRefreshKeys = (keys) ->
     matchingNodes = []
--- a/lib/assets/javascripts/turbograft/turbolinks.coffee
+++ b/lib/assets/javascripts/turbograft/turbolinks.coffee
@@ -121,23 +121,22 @@ class window.Turbolinks
     for k,v of options.headers
       xhr.setRequestHeader k, v
 
-    xhr.onload = ->
-      if xhr.status >= 500
-        Turbolinks.fullPageNavigate(url)
-      else
-        Turbolinks.loadPage(url, xhr, options)
-      xhr = null
-
-    xhr.onerror = ->
-      # Workaround for sinon xhr.abort()
-      if xhr.statusText == "abort"
+    new Promise (resolve) => 
+      xhr.onload = ->
+        if xhr.status >= 500
+          Turbolinks.fullPageNavigate(url)
+        else
+          resolve(Turbolinks.loadPage(url, xhr, options))
         xhr = null
-        return
-      Turbolinks.fullPageNavigate(url)
 
-    xhr.send()
+      xhr.onerror = ->
+        # Workaround for sinon xhr.abort()
+        if xhr.statusText == "abort"
+          xhr = null
+          return
+        Turbolinks.fullPageNavigate(url)
 
-    return
+      xhr.send()
 
   @loadPage: (url, xhr, options = {}) ->
     triggerEvent 'page:receive'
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -121,23 +121,22 @@ class window.Turbolinks
...
         xhr = null
-        return
-      Turbolinks.fullPageNavigate(url)
 
-    xhr.send()
+      xhr.onerror = ->
+        # Workaround for sinon xhr.abort()
+        if xhr.statusText == "abort"
+          xhr = null
+          return
```
lemonmade: Should probably reject here? Seems like this path leads to a never-ending promise

test/javascripts/turbolinks_test.coffee
```diff
@@ -389,6 +389,17 @@ describe 'Turbolinks', ->
         assert.include(testDocument.body.textContent, 'Hi bob')
         done()
 
+    it 'returns a promise resolved with a list of nodes that are new (freshly replaced)', (done) ->
```
lemonmade: Seems like we could use a few more tests for other codepaths that should resolve as promises with the nodes.

test/javascripts/turbolinks_test.coffee
```diff
@@ -389,6 +389,17 @@ describe 'Turbolinks', ->
...
+      visit(url: 'noScriptsOrLinkInHead', options: {partialReplace: true, onlyKeys: ['turbo-area']})
+        .then (nodes) ->
+          assert.instanceOf(nodes, Array)
+          assert.lengthOf(nodes, 1)
+          node = nodes[0]
+
+          assert.equal('turbo-area', node.id)
+          assert.equal('turbo-area', node.getAttribute('refresh'))
+          done()
```
lemonmade: Mocha supports just returning a promise, and it waits for its fulfillment to finish the test. Given that Coffeescript is implicitly returning the `visit()` promise, this `done()` probably isn't actually doing anything.

lib/assets/javascripts/turbograft/turbolinks.coffee
TheMallen: true.

test/javascripts/turbolinks_test.coffee
TheMallen: 👍

test/javascripts/turbolinks_test.coffee
TheMallen: Sounds fair. I'll test the reject flow as well.

lib/assets/javascripts/turbograft/turbolinks.coffee
TheMallen: I had skipped the flows that ended with a full page refresh since your js context will die anyway, but maybe we want to reject on those as well just for clarity?

qq99: Do you need to resolve with a copy of the XHR (here, and above)? IIRC, rejecting here would cause the promise `.fail` methods to be called, but people might want to treat a real failure (HTTP 500+) differently from this abort (e.g., don't flash an error message or whatever they're doing).  People might want to inspect the XHR (I've written some code that checks the status for abort and behaves differently) for `xhr.statusText == 'abort'` and react accordingly.

This path would occur if you were to call `Page.refresh` twice in quick succession right (and cared to bind to the promise)?

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -121,23 +121,24 @@ class window.Turbolinks
...
-      xhr = null
-
-    xhr.onerror = ->
-      # Workaround for sinon xhr.abort()
-      if xhr.statusText == "abort"
+    new Promise (resolve, reject) =>
+      xhr.onload = ->
+        if xhr.status >= 500
+          Turbolinks.fullPageNavigate(url)
+          reject()
```
qq99: I agree that HTTP500+ should be `reject`, but what about 400+?  I think those are kinda fail-y too, but we just `resolve` them in this implementation.  Perhaps you want another code path:

```
if 500+
  fullNavigate & reject
if 400~499
  loadPage & reject
else
  loadPage & resolve
```
?

Not 100% sure but figured I'd add a comment about it

lib/assets/javascripts/turbograft/turbolinks.coffee
```diff
@@ -202,7 +203,7 @@ class window.Turbolinks
       triggerEvent 'page:change'
       triggerEvent 'page:update'
 
-    return
+    return [].slice.call(body.children)
```
qq99: New behaviour here, who consumes this?  Is anyone relying on `undefined` return value?

What follows is general musing and not specifically related, but from what I can glean from re-reading this path, it seems we:

1. `refreshNodes(getNodesWithRefreshAlways(), body)` : old document has `refresh-always` nodes from new document swapped in
2. `persistStaticElements(body)` : we fetch nodes from `static` nodes from old document, then place them in new document's body (makes sense, maintains static-ness)
3. **Either** a.) `refreshAllExceptWithKeys(options.exceptKeys, body) : we refresh a bunch of nodes in new document's body, except a few user marked not to refresh or b.) `deleteRefreshNeverNodes(body) : we remove `refresh-never` nodes from new document body
4. Finally, we just replace old document's body with new document's body.

Step 4 makes step 1 seem redundant; why modify old document body if we're just going to replace it later? Perhaps step 1 is only useful for listeners listening for the event `page:before-partial-replace` (called within step1)?  It's tricky to even get a good feel for specifically what's happening to which document from these lines.  Think I should open an issue to refactor and make this more clear?

In any case, I feel like the return value from this function is meant to represent "only the list of nodes that have changed", but with the concept of static nodes persisted from old body potentially being within the children, I guess I'm at a weird situation where, "technically accurate" but maybe a bit misleading?  `body.children` could include `tg-static` nodes right?  But I guess more weirdly, `body.children` could be all new nodes, but some of their descendants could be old nodes.  Not sure if we care to make that distinction.  Maybe why we never returned anything before?

lib/assets/javascripts/turbograft/turbolinks.coffee
TheMallen: Resolving with the xhr makes sense to me.

lib/assets/javascripts/turbograft/turbolinks.coffee
TheMallen: I think resolving with the XHR makes the most sense for 400 series. That way we have a 1:1 correspondence of resolution and TG completing the navigation, and developers can manually check the status if they need to fork on 4XXs.

lib/assets/javascripts/turbograft/turbolinks.coffee
TheMallen: You're right, step 1 seems silly in that case. 

I'd like to see a refactor for sure.

The return value was changed because it was strange that it returned nothing when you changed all the nodes, I think. Memory is fuzzy after vacation. If that's the case I should have moved the change to another PR. I think maybe I was writing tests for the promise behaviour and ran into issues with inspecting the outcome from TG's apis in tests that dealt with full body changes?

Anyway, I'm cool with moving it to another PR if we want so we can talk about it in more detail, or dropping it.

qq99: I think it can't really hurt
I'd be surprised if anyone was expecting `undefined` or `null` return value here
I guess one could argue it's doing extra computation

TheMallen:
We can't keep this always asynchronous without massively breaking BC, so we can't really make it always return a promise.

PR #176, opened 2017-08-10
Add support for tg-remote-nopushstate, letting user choose whether or not to update URL after tg-remote redirects

qq99:
This PR allows one to add `data-tg-remote-nopushstate` to a `data-tg-remote` element.  In doing so, if the remote does a HTTP302 redirect, TG will not update the browser URL state.

The default throughout TG is `true` for `updatePushState`.  So, here, you opt in to turn it off for the 1 request

Take 2 of https://github.com/Shopify/turbograft/pull/156


qq99:
cc @Shopify/tnt

PR #179, opened 2017-09-26
🛠️ Make it shippable

TheMallen:
## Summary
This PR aims to make TG shippable again.

## The Problem
We've been unable to deploy the latest version of TG in our apps because we relied on some quirky behaviour. Namely, we have multiple call-sites that pass an XHR into `page.refresh` which already has listeners on it, intending those listeners to be invoked after the replace has occurred. The problem arises because changes to loadPage's implementation meant that some of these (which resulted in redirects) were (correctly) detected to not be true partial replaces (they redirected to other routes) and ideally should have head asset tracking.  Since head asset tracking needs to be asynchronous, the code was rendered broken.

TG does have an api for doing things after a particular refresh, with the `callback` option, but these weird cases are very hard to grep for and can be convoluted to switch over.

## The Solution
To make this shippable, we make it so that:
- redirects no longer override `partialReplace: false`
- `page.refresh(xhr, {partial: true})` are always synchronous

## 🎩  instructions
- point your project at your local copy of this branch with `path: 'path/to/turbograft'` in your gemfile
- navigate about, complete user flows, go to places that have redirects
- observe whether things break

## Closing
It's not ideal, but with our eventual goal being to switch over to our react stack imo it's best to just get this shippable so we can fix bugs (and / or add any necessary features) in the interim.

Feel free to ping me on slack for more context.

### DIFF
--- a/lib/assets/javascripts/turbograft/response.coffee
+++ b/lib/assets/javascripts/turbograft/response.coffee
@@ -1,11 +1,11 @@
 class TurboGraft.Response
   constructor: (@xhr, intendedURL) ->
     if intendedURL && intendedURL.withoutHash() != @xhr.responseURL
-      @redirectedTo = @xhr.responseURL
+      redirectedTo = @xhr.responseURL
     else
-      @redirectedTo = @xhr.getResponseHeader('X-XHR-Redirected-To')
+      redirectedTo = @xhr.getResponseHeader('X-XHR-Redirected-To')
 
-    @finalURL = @redirectedTo || intendedURL
+    @finalURL = redirectedTo || intendedURL
 
   valid: -> @hasRenderableHttpStatus() && @hasValidContent()
 
@@ -22,12 +22,6 @@ class TurboGraft.Response
       contentType.match(/^(?:text\/html|application\/xhtml\+xml|application\/xml)(?:;|$)/)
     else
       throw new Error("Error encountered for XHR Response: #{this}")
-
-  redirectedToNewUrl: () ->
-    Boolean(
-      @redirectedTo &&
-      @redirectedTo != TurboGraft.location()
-    )
 
   toString: () ->
     "URL: #{@xhr.responseURL}, " +
--- a/test/javascripts/turbolinks_test.coffee
+++ b/test/javascripts/turbolinks_test.coffee
@@ -150,6 +150,30 @@ describe 'Turbolinks', ->
         assert.calledOnce(resetScrollStub)
         done()
 
+  describe 'loadPage', () ->
+    PAGE_CONTENT = 'Oh hi, mark!'
+
+    it 'is synchronous when a partial replace', ->
+      startFromFixture('noScriptsOrLinkInHead')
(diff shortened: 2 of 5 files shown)


GoodForOneFare:
Work in progress - I'll update when I finish each test run.

| Action | Result | Chrome | IE 11 | Edge | Safari 11 |
| ----- | ----- | ----- | ----- | ----- | ----- |
| In a new draft order, click _Mark as Paid_ | Draft order saved | 💚 | - | - | - |
|  | Redirects to draft order with id in URL | 💚 | - | - | - |
|  | CC modal appears | 💚 | - | - | - |
| In an existing draft order, click _Mark as Paid_ | CC modal appears| 💚 | - | - | - |
| Navigate to the theme editor | Page refreshes | 💚 | - | - | - |
|  | Navigating back via a link refreshes the page | 💚 | - | - | - |
|  | Themes page is fully styles | 💚 | - | - | - |
| Navigated to the theme editor | Page refreshes | 💚 | - | - | - |
|  | Navigating back via browser back button refreshes the page | 💚 | - | - | - |
| On Home, find the _Your Online Store isn't open yet_ card and click _Remove password page_ card | Transitions to settings page _without_ a page refresh | 💚 | - | - | - |
|  | Scrolls the page down to reveal the shop password section ✨  | 💚 | - | - | - |
| Clear Shopify cookies, and navigate via a sidebar link | Redirects to a fully styled login page | 💚 | - | - | - |
| Go to theme editor, clear cookies, and navigate using the browser back button | Redirected to a fully styled login page | 💚 | - | - | - |

lib/assets/javascripts/turbograft/response.coffee
```diff
@@ -23,12 +23,6 @@ class TurboGraft.Response
     else
       throw new Error("Error encountered for XHR Response: #{this}")
 
-  redirectedToNewUrl: () ->
-    Boolean(
-      @redirectedTo &&
```
GoodForOneFare: Does `@redirectedTo` need to be in class scope, now?

test/javascripts/turbolinks_test.coffee
```diff
@@ -150,6 +150,30 @@ describe 'Turbolinks', ->
...
+  describe 'loadPage', () ->
+    PAGE_CONTENT = 'Oh hi, mark!'
+
+    it 'is synchronous when a partial replace', ->
+      startFromFixture('noScriptsOrLinkInHead')
+      xhr = new sinon.FakeXMLHttpRequest()
+      xhr.open('POST', '/my/endpoint', true)
+      xhr.respond(200, {'Content-Type':'text/html'}, PAGE_CONTENT)
+
+      loadPageReturn = Turbolinks.loadPage(null, xhr, {partialReplace: true})
```
qq99: assignment of `loadPageReturn =` not necessary it looks like

lib/assets/javascripts/turbograft/response.coffee
TheMallen: good point, it can just be an implementation detail of `finalURL`

PR #188, opened 2023-09-11
Remove coffeescript from this library

rafaelfranca:
Don't force the coffee-rails library as dependencies of applications.


rafaelfranca:
Tests are green and diff of resulting JavaScript doesn't show any big differences.

lib/assets/javascripts/turbograft/click.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+// The Click class handles clicked links, verifying if Turbolinks should
+// take control by inspecting both the event and the link. If it should,
+// the page change process is initiated. If not, control is passed back
+// to the browser for default functionality.
+window.Click = class Click {
+  static installHandlerLast(event) {
+    if (!event.defaultPrevented) {
+      document.removeEventListener('click', Click.handle, false);
+      return document.addEventListener('click', Click.handle, false);
```
paracycle, line 15: I think this can be omitted, especially since `document.addEventListener` [does not even return anything](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#return_value).
```suggestion
      document.addEventListener('click', Click.handle, false);
```

lib/assets/javascripts/turbograft/click.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  _extractLink() {
+    let link = this.event.target;
+    while (!!link.parentNode && (link.nodeName !== 'A')) { link = link.parentNode; }
+    if ((link.nodeName === 'A') && (link.href.length !== 0)) { return this.link = new Link(link); }
+  }
+
+  _validForTurbolinks() {
+    return (this.link != null) && !(this.link.shouldIgnore() || this._nonStandardClick());
```
paracycle, line 40: I think:
```suggestion
    return (this.link != null) && !this.link.shouldIgnore() && !this._nonStandardClick();
```
would be easier to read in JS.

lib/assets/javascripts/turbograft/component_url.js
```diff
@@ -0,0 +1,56 @@
+/* The ComponentUrl class converts a basic URL string into an object
+ * that behaves similarly to document.location.
+ *
+ * If an instance is created from a relative URL, the current document
+ * is used to fill in the missing attributes (protocol, host, port).
+ */
+window.ComponentUrl = (function() {
```
paracycle, line 7: Why is this using an IIFE instead of a `class` defined class like `Click` does?

lib/assets/javascripts/turbograft.js
```diff
@@ -0,0 +1,37 @@
+//= require_self
+//= require_tree ./turbograft
+
+if (window.TurboGraft == null) { window.TurboGraft = { handlers: {} }; }
+
+TurboGraft.tgAttribute = function(attr) {
+  if (attr.slice(0, 3) === 'tg-') {
+    return `data-${attr}`;
+  } else {
+    return `data-tg-${attr}`;
```
gmcgibbon, line 10: https://caniuse.com/?search=template%20string

FYI This isn't supported by all browsers, and neither is `class` and some of ES6 features we're using. Though, given our preference to use import maps, I imagine we don't care about these edge cases.

lib/assets/javascripts/turbograft/csrf_token.js
```diff
@@ -0,0 +1,29 @@
...
+      object.token = tag.getAttribute('content');
+    }
+
+    return object;
+  }
+
+  static update(latest) {
+    const current = this.get();
+    if ((current.token != null) && (latest != null) && (current.token !== latest)) {
+      return current.node.setAttribute('content', latest);
```
paracycle, line 26: Again, this is unnecessary, since `setAttribute` [does not return anything](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute#return_value).
```suggestion
      current.node.setAttribute('content', latest);
```

lib/assets/javascripts/turbograft/component_url.js
gmcgibbon: Agree, this seems odd. Classes are also not supported by some browsers, but maybe that's ok: https://caniuse.com/?search=class

test/teaspoon_env.rb
```diff
@@ -13,7 +13,7 @@
 
   config.suite do |suite|
     suite.use_framework :mocha
-    suite.matcher = "{test/javascripts,app/assets}/**/*_test.{js,js.coffee,coffee,coffee.erb}"
+    suite.matcher = "{test/javascripts,app/assets}/**/*_test.js"
```
gmcgibbon, line 16:
```suggestion
    suite.matcher = "{test/javascripts,app/assets}/**/*Test.js"
```
I would say using camelCase is more of a JS standard, but this is a nitpick.

lib/assets/javascripts/turbograft/component_url.js
rafaelfranca: That is what coffeescript generated. If I use class, tests start to fail, no idea why, so I just kept the code as coffeescript generates.

lib/assets/javascripts/turbograft.js
rafaelfranca: ES6 is compiled to ES5 by sprockets in the final result. Those assets are never returned to users as is.

lib/assets/javascripts/turbograft/csrf_token.js
gmcgibbon: I would imagine this happens a lot with CoffeeScript's implicit returns.

lib/assets/javascripts/turbograft/component_url.js
gmcgibbon: It might be related to function binding, but it hard to guess without output. We could also open an issue for this and do it later.

rafaelfranca: I see no usage of binding in coffeescript, and the code on this class is so simple that I don't see how that would influence the behavior to the point that tests would fail.

rafaelfranca:
This repository will probably not evolve further. All code in our apps that use it should at this point be deleted. All code that is kept, should just use Turbo. There is no maintainer for this library, no CI. I'm just doing this migration because I want to remove CoffeeScript from the applications we are using this gem.

I don't see need to add eslint, or anything to improve this code further given that.

lib/assets/javascripts/turbograft/link.js
```diff
@@ -0,0 +1,67 @@
...
+        Link.HTML_EXTENSIONS.push(extension);
+      }
+      return Link.HTML_EXTENSIONS;
+    };
+
+    function Link(link) {
+      if (link.constructor === Link) {
+        return link;
+      }
+      Link.__super__.constructor.call(this, link.href, link);
```
paracycle, line 29: Ok, I figured out that this line needs to be changed to make all the `class` syntax to work:
```suggestion
      super(link.href, link);
```
Otherwise, JS complains that a constructor was not called with `new`.

lib/assets/javascripts/turbograft/component_url.js
paracycle: I think I fixed this in my last commit.

lib/assets/javascripts/turbograft/link.js
```diff
@@ -0,0 +1,67 @@
+(function() {
+  var extend = function(child, parent) { for (var key in parent) { if (hasProp.call(parent, key)) child[key] = parent[key]; } function ctor() { this.constructor = child; } ctor.prototype = parent.prototype; child.prototype = new ctor(); child.__super__ = parent.prototype; return child; },
+    hasProp = {}.hasOwnProperty,
+    slice = [].slice;
+
+  /* The Link class derives from the ComponentUrl class, but is built from an
+   * existing link element.  Provides verification functionality for Turbolinks
+   * to use in determining whether it should process the link when clicked.
+   */
+  window.Link = (function(superClass) {
```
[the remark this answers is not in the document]
paracycle: I think I fixed this in my last commit.

PR #189, opened 2023-09-14
Define the field as a class property

rafaelfranca:


paracycle (approved):
Thank you for fixing this!

